PR #8, opened 2016-02-26
fall over early if dials.refine fails

Anthchirp:
example code review


graeme-winter:
Looks good at a quick glance (like the tool)

PR #130, opened 2017-06-12
DIALS 1.6.1

Anthchirp:


Anthchirp:
xia2 regression tests currently failing.

PR #252, opened 2018-09-25
Drop libtbx testing

Anthchirp:
Discuss


graeme-winter:
So in place of 

```
libtbx.run_tests_parallel \
run_in_tmp_dir=true \
module=xia2 module=cctbx module=rstbx module=dxtbx module=dials \
module=dials_regression \
nproc=6
```

I need to have something which also steps into the xia2 directory and summons pytest? No unreasonable, but I will need to update my scripts.

Anthchirp:
You can run
* ```libtbx.pytest -ra -n auto $path_to_xia2_folder``` for tests not involving regression data (very fast)
* ```libtbx.pytest -ra -n auto --regression $path_to_xia2_folder``` for all tests (very opposite)

This works from anywhere. However you do get nicer output with
* ```pushd $path_to_xia2_folder; libtbx.pytest -ra -n auto; popd```
* ```pushd $path_to_xia2_folder; libtbx.pytest -ra -n auto --regression; popd```

because you get the benefit of shorter file paths.

PR #255, opened 2018-09-28
Probplots in report

### LINKED ISSUE #234: Error analysis: add to xia2 report

benjaminhwilliams:
Dropped the ball on this a little while ago.  Picking it up again.

Questions before this pull request can be accepted:
1. Should this information be in [`xia2/command_line/report.py`](https://github.com/xia2/xia2/blob/probplots_in_report/command_line/report.py) or would it be better for it to live in [`dials/command_line/report`](https://github.com/dials/dials/blob/probplots_in_report/command_line/report.py)?  See [parallel pull request in `dials/dials`](https://github.com/dials/dials/pull/622).
2. Because I started without a clear idea of the answer to (1), the nuts and bolts of this are currently living in [`dials/util/intensity_explorer.py`](https://github.com/dials/dials/blob/probplots_in_report/util/intensity_explorer.py).  If the answer to (1) is `xia2/command_line/report.py`, then I should move `intensity_explorer.py` to this repo, to keep the imports internal.  Where should it live?
3. In fact, even a satisfactory answer to (2) doesn't remove a dependency on DIALS, since `intensity_explorer.py` creates a `dials_array_family_flex_ext.reflection_table` anyway.  Is that a problem?

Closes #234.


Anthchirp:
3. is definitely not a problem, xia2 depends on quite a few bits of DIALS. The other way around would be more problematic.

Anthchirp:
(should be squashed on merge)

benjaminhwilliams:
OK, so the only outstanding question is (1), i.e. whether this information belongs more in the xia2 report or the DIALS report (or both).  My inclination is towards the DIALS report, since this information is probably more of interest to the more expert user.  But I defer to @Anthchirp & @graeme-winter, since this is more of a design decision.

benjaminhwilliams:
In answer to my question as to whether this pull request or [this parallel DIALS pull request](https://github.com/dials/dials/pull/622) should be merged, it has been decided to merge both.

jbeilstenedmands:
I notice in xia2 that there is the xia2-report.html and xia2.html : xia2.html seems to include the same graphs as the report except for the new intensity plots, should these be added there as well?

PR #290, opened 2019-04-02
Migrate plotting code from xia2 to dials

jbeilstenedmands:
Plots have been moved to dials, to allow them to be used for dials.report and the output of other dials programs. Moved are batch plots, resolution dependent plots, xtriage plots, statistics tables.

The plotting classes and functions can be called from xia2.report without significant changes. All existing phil scope options are preserved.

Thought i'd put this in as a pull request so that it can be checked over as I'm removing a fair chunk of code from xia2. The code has been moved to dials.report.plots, dials.report.analysis and dials.util.batch_handling, with some refactoring in parts to help dials.report and dials.scale use them for html output.
I've tested generating the xia2.report and everything looks the same as before as far as I can tell.
I'll probably move some more plots over later in the week.


jbeilstenedmands:
I have just realised the fact that xia2.multiplex will also use the report code and will now be broken. Will fix this and push another commit.

jbeilstenedmands:
@rjgildea do you wish for the xtriage plots to be present in the xia2.multiplex report? I notice there's a call to the Wilson, l_test and cumulative_intensity plots in multi_crystal_analysis code but these are not generated due to the structure of xia2_report_base. They now will be plotted due to the changes in this PR, but I can remove them if they aren't intended to be there.

rjgildea:
Yes, I think these plots were probably supposed to be included in the report.

PR #295, opened 2019-04-23
pickle to msgpack changes

rjgildea:
This makes the necessary changes to xia2 now that msgpack is the default reflection file format in dials (dials/dials#743). All tests pass with `$ pytest --regression-full -n auto` apart from 2 tests which fail tolerance tests (I think these were failing before the msgpack changes were merged in to dials master).

### DIFF
--- a/Modules/Scaler/rebatch.py
+++ b/Modules/Scaler/rebatch.py
@@ -27,6 +27,7 @@ def compact_batches(batches):
         for k, g in groupby(enumerate(batches), lambda i_x: i_x[0] - i_x[1])
     ]
 
+import time
 
 def rebatch(
     hklin,
@@ -42,6 +43,7 @@ def rebatch(
 ):
     """Need to implement: include batch range, exclude batches, add N to
     batches, start batches at N."""
+    t0 = time.time()
     if include_range is None:
         include_range = []
     if exclude_range is None:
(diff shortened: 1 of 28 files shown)


graeme-winter:
Wilco

command_line/report.py
```diff
@@ -1160,14 +1160,13 @@ def z_vs_I_over_sigma(self):
...
+        import iotbx.mtz
 
         self.unmerged_mtz = unmerged_mtz
         self.params = params
 
-        reader = any_reflection_file(unmerged_mtz)
-        assert reader.file_type() == "ccp4_mtz"
-        arrays = reader.as_miller_arrays(merge_equivalents=False)
+        self.mtz_object = iotbx.mtz.object(unmerged_mtz)
+        arrays = self.mtz_object.as_miller_arrays(merge_equivalents=False, auto_anomalous=True)
```
Anthchirp: Changes in this file are unrelated?

Modules/Scaler/rebatch.py
```diff
@@ -122,12 +141,28 @@ def rebatch(
...
+
     new_batches = (min(batches) + offset, max(batches) + offset)
 
+    t9 = time.time()
+
+    print(t1 - t0)
+    print(t2 - t1)
+    print(t3 - t2)
+    print(t4 - t3)
+    print(t5 - t4)
```
Anthchirp: accidental add?

command_line/report.py
rjgildea: It turns out pull requests are useful 😉

PR #296, opened 2019-04-26
LGTM test

Anthchirp:
apparently they now have a github app, so let's have a look


Anthchirp:
maybe don't merge this then.

PR #297, opened 2019-04-30
Partial revert of msgpack changes

rjgildea:
Revert only the filenames and writing, keeping the changes for reading as they will work independent of reflection file format. Revisit the output file format once msgpack performance issues with large reflection files have been addressed.


ndevenish:
Probably worth squashing into one commit to make it easier to wholesale revert again when required.

PR #301, opened 2019-05-08
Change the default scaler for pipeline=dials

graeme-winter:
Old behaviour dials-legacy - default is dials which is the same as dials-full; dials-full left to avoid crashes though should deprecate

### DIFF
--- a/Handlers/CommandLine.py
+++ b/Handlers/CommandLine.py
@@ -560,10 +560,13 @@ class _CommandLine(object):
             indexer, refiner, integrater, scaler = "dials", "xds", "xdsr", "xdsa"
         elif settings.pipeline == "dials":
             Debug.write("DIALS pipeline selected")
-            indexer, refiner, integrater, scaler = "dials", "dials", "dials", "ccp4a"
-        elif settings.pipeline == "dials-full":
-            Debug.write("DIALS-FULL pipeline selected")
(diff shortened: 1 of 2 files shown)


graeme-winter:
Aware that tests will need updating. @jbeilstenedmands you happy with this happening?

jbeilstenedmands:
Yes happy with this happening. Will probably require some extra attention on the beamlines when the switch is made to catch any leftover bugs - when do you plan to merge this in?

graeme-winter:
Discussion topic - though beamlines don’t run off master usually so probably OK…

Anthchirp:
Happy for this to go in. Would be for 2.0 release only, so won't hit beamlines for a while.

graeme-winter:
+1 to additional changes

rjgildea:
Could "dials-legacy" just be "dials-aimless" which is a more usefully descriptive name?

graeme-winter:
> Could "dials-legacy" just be "dials-aimless" which is a more usefully descriptive name?

Make it so 🙂

Anthchirp:
Happy with this going into the release notes / changelog? (if not feel free to change)
https://github.com/xia2/xia2/pull/301/commits/6a9d5bc80d9cb29a8a0a0be132722ef37fbb461e

jbeilstenedmands:
@Anthchirp has pointed out that dials-full is not producing the .sca output files (by design). @graeme-winter I guess we want to be making these?

PR #304, opened 2019-05-22
Don't use reserved names - type and list

jbeilstenedmands:
Unless there is some magic happening, we probably shouldn't be doing this? Could be hiding some issues perhaps.

### DIFF
--- a/Decorators/DecoratorFactory.py
+++ b/Decorators/DecoratorFactory.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
-# DecoratorFactory.py
-#
 #   Copyright (C) 2006 CCLRC, Graeme Winter
 #
 #   This code is distributed under the BSD license, a copy of which is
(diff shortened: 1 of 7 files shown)


Wrappers/XDS/XDSIdxref.py
```diff
@@ -510,7 +510,7 @@ def continue_from_error(self):
 
             # get the highest symmetry "acceptable" solution
 
-            list = [
+            lst = [
```
rjgildea: Perhaps "items" might be better than "lst" here?

Handlers/Syminfo.py
```diff
@@ -65,11 +65,11 @@ def _parse_symop(self):
 
         for line in open(os.path.join(os.environ["CLIBD"], "symop.lib")).readlines():
             if line[0] != " ":
-                list = line.split()
-                index = int(list[0])
-                shortname = list[3]
+                lst = line.split()
```
rjgildea: Maybe prefer "tokens" instead of "lst"

Decorators/DecoratorFactory.py
```diff
@@ -25,16 +25,16 @@ class _DecoratorFactory(object):
     def __init__(self):
         self._type = "ccp4"
 
-    def Decorate(self, DriverInstance, type=None):
+    def Decorate(self, DriverInstance, typ=None):
```
rjgildea: I would prefer a more meaningful name than "typ", possibly "decorator_type" or "decorator_class"

Driver/DriverFactory.py
```diff
@@ -48,39 +48,39 @@ def __init__(self):
...
+            raise RuntimeError("unimplemented driver class: %s" % typ)
 
-        self._driver_type = type
+        self._driver_type = typ
 
     def get_driver_type(self):
         return self._driver_type
 
-    def Driver(self, type=None):
+    def Driver(self, typ=None):
```
rjgildea: Similarly "driver_type" might be better

Driver/ClusterDriverFactory.py
```diff
@@ -38,30 +38,30 @@ def __init__(self):
...
+        if not typ in self._implemented_types:
+            raise RuntimeError("unimplemented driver class: %s" % typ)
 
-        self._driver_type = type
+        self._driver_type = typ
 
         return
 
-    def Driver(self, type=None):
+    def Driver(self, typ=None):
```
rjgildea: Similarly "driver_type" might be better

PR #310, opened 2019-06-21
Rename output files

dagewa:
* I have not done a --full-regression test yet
* I am unsure if Driver/test_timing.py also requires changes
* Handlers/Phil.py has input.json=. Perhaps this should be input.expt=?

### DIFF
--- a/command_line/overload.py
+++ b/command_line/overload.py
@@ -93,7 +93,7 @@ def build_hist(nproc=1):
     if len(sys.argv) >= 2 and sys.argv[1].startswith("nproc="):
         nproc = int(sys.argv[1][6:])
         sys.argv = sys.argv[1:]
-    if len(sys.argv) == 2 and sys.argv[1][-5:] in (".expt", ".json"):
+    if len(sys.argv) == 2 and sys.argv[1].endswith((".expt", ".json")):
         from dxtbx.model.experiment_list import ExperimentListFactory
 
         experiments = ExperimentListFactory.from_json_file(sys.argv[1])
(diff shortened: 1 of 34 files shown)


command_line/to_shelx.py
```diff
@@ -192,7 +192,7 @@ def to_shelx(hklin, prefix, compound="", options=None):
     unit_cell_esds = None
     cell_data = None
     if options and options.cell:
-        if options.cell.endswith(".json"):
+        if options.cell[-5:] in (".expt", ".json"):
```
Anthchirp:
```suggestion
        if options.cell.endswith((".expt", ".json"))
```

command_line/overload.py
```diff
@@ -93,7 +93,7 @@ def build_hist(nproc=1):
     if len(sys.argv) >= 2 and sys.argv[1].startswith("nproc="):
         nproc = int(sys.argv[1][6:])
         sys.argv = sys.argv[1:]
-    if len(sys.argv) == 2 and sys.argv[1].endswith(".json"):
+    if len(sys.argv) == 2 and sys.argv[1][-5:] in (".expt", ".json"):
```
Anthchirp:
```suggestion
    if len(sys.argv) == 2 and sys.argv[1].endswith((".expt", ".json")):
```

command_line/to_shelx.py
dagewa: Nice, I did not know `endswith` can take multiple arguments

Anthchirp:
`Driver/test_timing.py` does not need to be changed. Command line values in there are dummy values that come out of a real run.

PR #317, opened 2019-07-17
Multiplex report refactor

rjgildea:
* Some refactoring of report code to enable more general re-use.
* Rename multiplex report to xia2.multiplex.html
* Update multiplex report to have merging stats/plots for each scaled cluster
* Add cluster comparision graphs to multiplex html


rjgildea:
Minimal example case to perform scaling on multiple clusters (all clusters with > 50% completeness):
```
$ xia2.multiplex $(dials.data get -q multi_crystal_proteinase_k)/* min_completeness=0.5
```

Add `max_clusters=n` parameter to limit to top `n` clusters.

graeme-winter:
Reviewing now 👍

Issue #267, opened 2018-12-05
Replace xia2 Chatter/Debug streams with proper Python logging calls


rjgildea:

Anthchirp:
* [ ] The next step would be to use setup_logging() in all xia2 dispatchers, at which point we can remove the reconfigure_streams_to_logging() function and roll its changes into the Chatter/Debug/Stdout classes.
* [ ] From there on we can use logging as normal throughout xia2 and have it co-exist with the stream classes for a while.
* [ ] Finally we can deprecate and remove the stream classes.

PR #320, opened 2019-07-26
Start replacing Chatter/Debug streams with logging

Anthchirp:
This commit adds an adapter function that replaces the backend file
operations in Chatter/Debug/Stdout streams with calls to Python logging.
With this we can support code that uses Chatter/Debug as well as code
that uses logging.

NB:
* if Chatter/Debug should use the logging backend then this should be
  set up before the first .write() call happens. Otherwise the output
  file will be created.
* all messages coming in via Chatter/Stdout have loglevel info, all
  messages coming in via Debug have loglevel debug.
* the adapter only goes one way. If a command that does not set up
  the logging backend calls code that uses logging then that output
  does not go into the log.

Adding this would be the first step towards #267.
The next step would be to use `setup_logging()` in all xia2 dispatchers, at which point we can remove the `reconfigure_streams_to_logging()` function and roll its changes into the `Chatter`/`Debug`/`Stdout` classes.
From there on we can use logging as normal throughout xia2 and have it co-exist with the stream classes for a while.
Finally we can deprecate and remove the stream classes.

### DIFF
--- a/Handlers/Streams.py
+++ b/Handlers/Streams.py
@@ -276,5 +276,39 @@ def setup_logging(logfile=None, debugfile=None, verbose=False):
             logger.setLevel(logging.DEBUG)
 
 
+def reconfigure_streams_to_logging():
+    "Modify xia2 Chatter/Debug objects to use python logging"
+
+    def logger_file(loggername, level=logging.INFO):
+        "Returns a file-like object that writes to a logger"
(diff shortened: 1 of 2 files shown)


graeme-winter:
Running test to see if this has broken anything right now - looks like ... no, not broken and the console output seems ~ identical - at least modulo the other changes so will squash merge.

PR #325, opened 2019-08-16
Deprecation warnings

benjaminhwilliams:
While working in `DialsIndexer`, I noticed a deprecation warning, so I did a little semi-automatic tidy-up.  This PR does the following, based on the advice in the deprecation warnings:
* Replace `dxtbx.model.experiment_list.ExperimentListDumper` and `dxtbx.serialize.dump` with `as_file` method of experiment list;
* Remove `xia2.Applications.xia2setup.save_experiments`, which doesn't seem to be used anywhere;
* Remove some unused imports and variables to satisfy flake8.

Any objections?

There are corresponding PRs to follow in dials, dials_scratch & dxtbx.

### DIFF
--- a/Modules/Indexer/DialsIndexer.py
+++ b/Modules/Indexer/DialsIndexer.py
@@ -135,14 +135,12 @@ class DialsIndexer(Indexer):
 
     def Refine(self):
         refine = _Refine()
-        params = PhilIndex.params.dials.refine
+        params = PhilIndex.params.dials
         refine.set_working_directory(self.get_working_directory())
         refine.set_scan_varying(False)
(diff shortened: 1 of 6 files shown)


Modules/Indexer/DialsIndexer.py
```diff
@@ -487,7 +485,8 @@ def _index(self):
                 elif nref_3d is not None:
                     indexer = indexer_fft3d
                 else:
-                    raise RuntimeError(e)
+                    # Flake8 doesn't recognise this as `e` from above.
```
benjaminhwilliams: Aaaaaah, this is different behaviour in Python 3, the value of `e` isn't retained from the `except` statement above.

Anthchirp:
lgtm

PR #329, opened 2019-09-16
Updates to avoid excessive calls to dials.export

jbeilstenedmands:
Update calls to resolutionizer to match new dials code - use native .refl and .expt files.

Don't export mtz by default in dials.scale as not necessary.

When constructing DialsIntegrater, set option to not export mtz files if dials-full (==dials) pipeline as not needed for subsequent processes.

### DIFF
--- a/Modules/Scaler/DialsScaler.py
+++ b/Modules/Scaler/DialsScaler.py
@@ -508,14 +508,6 @@ pipeline=dials (supported for pipeline=dials-aimless).
         self._scalr_scaled_reflection_files["mtz_unmerged"] = {}
         self._scalr_scaled_reflection_files["mtz"] = {}
 
-        # Set the unmerged mtz output filename - need for merging stats calc.
-        scaled_unmerged_mtz_path = os.path.join(
-            self.get_working_directory(),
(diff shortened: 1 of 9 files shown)


Modules/Scaler/DialsScaler.py
```diff
@@ -567,9 +557,11 @@ def _scale(self):
         ### Calculate the resolution limit and set done False if applicable
 
         highest_suggested_resolution = self.assess_resolution_limits(
-            self._scaler.get_unmerged_reflection_file(),
-            user_resolution_limits,
+            hklin=None,  # self._scaler.get_unmerged_reflection_file(),
```
benjaminhwilliams: Is the commented parameter kept deliberately?

dagewa:
Just a note that at the DLS/CCP4 workshop some users were confused that they couldn't find the integrated.mtz in iSPyB. They wanted to scale together multiple MTZ files from different runs, so I suggested they take the scaled, unmerged MTZs.

rjgildea:
See https://github.com/xia2/xia2/issues/385

PR #340, opened 2019-10-22
Allow unicode output in HTML report

benjaminhwilliams:
Not intended for merge until after DIALS 2.0 release.


benjaminhwilliams:
I've been idly tinkering with this without giving it my full attention.  It won't fly in its current form.  I'll need to focus properly on making this work when I have more time.  Closing for now.

PR #342, opened 2019-10-23
Allow unicode output in HTML report

benjaminhwilliams:
This time, I think it works...

Not intended for merge until after DIALS 2.0 release.

Depends on dials/dials#983.


benjaminhwilliams:
Better articulated:
At present, if you put a unicode character in any `Chatter.write` (and presumably also `Chatter.debug`) call, the HTML report generation falls over with a `UnicodeEncodeError` (or maybe it was a `UnicodeDecodeError`, I've forgotten).  This fixes that, so now you can write all the unicode you want.

Μπορῶ νὰ φάω σπασμένα γυαλιὰ χωρὶς νὰ πάθω τίποτα.

benjaminhwilliams:
Following the merge of dials/dials#983, this too should be good to go.

benjaminhwilliams:
I tested on a Python 2 build, so I think we're good to go.  I'll let the last checks finish from my latest merge of commits from master, then push the green button if all is still well.

PR #344, opened 2019-10-28
Retire mosflm/2d pipeline

### LINKED ISSUE #222: "Retire" mosflm / 2d pipeline
Since I do not think this is much used and we want to move forward, need to retire to reduce amount of code to maintain. Asked CCP4; response was "if they want mosflm, use imosflm"

Obviously old versions of xia2 will be available too...

Anthchirp:
Closes #222


graeme-winter:
Scanned diffs

Assuming all the tests pass (which I will take as a given) then IMHO we're good to go. 

Appreciate this thank you - also - 4k lines of code is _a good thing_

PR #353, opened 2019-11-05
Python 3: lib2to3.fixes.fix_map

Anthchirp:
more xia2 refactoring work to run on Python 3


Anthchirp:
full regression tests pass now

PR #370, opened 2019-11-12
Update command line tool log streaming interface

Anthchirp:
Update all command line tools that import xia2 components to explicitly
create named log files and redirect all `Chatter()` and other stream
output to said log files.

Some commands that do not import from xia2 were not touched.

This is the first half of the first step of the list in #267


Anthchirp:
Side effects:
* `xia2.multiplex` may produce more output, as streams were switched off previously
* this PR retires the name `dev.xia2.strategy`, the script has been known as `xia2.strategy` for 3 years now

PR #374, opened 2019-11-14
Only use cc-half cutoff for all pipelines

jbeilstenedmands:
Currently, in the dials pipeline, when determining the resolution cutoff, the higher of the cc1/2 and isigma cutoffs (found by dials.resolutionizer) are used. In the dials-aimless pipeline, misigma is also used.
However, it is now widely accepted [1] that CC-half is the only real suitable metric for this purpose. It is known that the isigma cutoff is unreliable due to the unreliability of the sigmas, which are susceptible to the form of error modelling used in scaling. As I've been working on the dials.scale error model I have noticed the resolution cutoffs on isigma move around quite a bit, however the cchalf cutoffs remain the same. In addition, the dials.scale error modelling will probably continue to evolve over the next few years, making it annoying to always update the regression tests due to changing resolution limits (my recent dials commit will break the regression tests later today).

So I propose we just use CC-half for the dials pipeline, and could also consider moving over other pipelines too. Thoughts?

[1] https://doi.org/10.1016/j.sbi.2015.07.003

### DIFF
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -14,7 +14,7 @@ from __future__ import absolute_import, division, print_function
 from iotbx.phil import parse
 from libtbx.phil import interface
 
-phil_scope = parse(
+master_phil = parse(
     """
 general
   .short_caption = "General settings"
@@ -940,7 +940,7 @@ xia2.settings
 )
 
 # override default resolutionizer parameters
-phil_overrides = phil_scope.fetch(
+master_phil = master_phil.fetch(
     source=parse(
         """\
 xia2.settings {
@@ -952,8 +952,6 @@ xia2.settings {
 """
     )
 )
-
-master_phil = phil_scope.fetch(sources=[phil_overrides])
 
 PhilIndex = interface.index(master_phil=master_phil)
 
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -741,6 +741,14 @@ xia2.settings
       .type = float(value_min=0.0)
       .help = "High resolution cutoff."
       .short_caption = "High resolution cutoff"
+    use_isigma = False
+      .type = bool
+      .help = "Use an isigma resolution cutoff (value specified by resolution.isigma)"
+      .short_caption = "Use an isigma cutoff for resolution"
+    use_misigma = False
+      .type = bool
+      .help = "Use a mean isigma resolution cutoff (value specified by resolution.misigma)"
+      .short_caption = "Use a mean isigma cutoff for resolution"
     include scope dials.util.resolutionizer.phil_str
   }
   unify_setting = False
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -14,7 +14,7 @@ from __future__ import absolute_import, division, print_function
 from iotbx.phil import parse
 from libtbx.phil import interface
 
-master_phil = parse(
+phil_scope = parse(
     """
 general
   .short_caption = "General settings"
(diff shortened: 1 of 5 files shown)


rjgildea:
I think the concept is sound, but the proposed change would mean it would no longer be possible to use I/sigi even if you wanted to. I suggest either changing the defaults in:
https://github.com/dials/dials/blob/7c9ee2c0fe391fb9a326e911e22adddeb30df07f/util/resolutionizer.py#L250-L259
to `isigma=None; misigma=None`, or overriding them when we import the phil scope in Handlers/Phil.py:
https://github.com/xia2/xia2/blob/master/Handlers/Phil.py#L744

rjgildea:
If we think that in general `cc_half` only is the sensible approach for DIALS, then it makes sense to change the defaults in `dials.resolutionizer` to reflect this.

jbeilstenedmands:
Thanks for the feedback, I hadn't noticed the adoption of the resolutionizer phil scope, and it's not clear how/if this is plumbed in. I'll have a look so that we can integrate this into the resolution cutting if we want to go down this route.

rjgildea:
These lines shouldn't really be needed:
https://github.com/xia2/xia2/blob/4d8cfa802576bb710c0f3d4a40cb6266bf121cf5/Modules/Scaler/CommonScaler.py#L993-L996
as whether or not a particular resolution cutoff method is used depends on whether or not the relevant phil parameter is set to a value or `None`:
https://github.com/xia2/xia2/blob/4d8cfa802576bb710c0f3d4a40cb6266bf121cf5/Modules/Scaler/CommonScaler.py#L1007-L1030

jbeilstenedmands:
From testing, it seems like those lines do have an effect (as Merger sets the params to None for resolutionizer by default). Therefore I propose adding two phil options, use_isigma and use_misigma, which are False by default. If set to True, then the default values from resolutionizer are picked up. With this setup it also make sense to be consistent across pipelines, so I have updated so that all take these parameters as options.

rjgildea:
I don't understand why it is necessary to have two additional parameters. Surely setting `use_isgima=False` is equivalent to doing `isigma=None`? So we effectively have two parameters to do the same thing...

graeme-winter:
Agree that limit=None is identical to use_limit=no and affords less chance of ambiguity

Modules/Scaler/CommonScaler.py
```diff
@@ -990,9 +984,9 @@ def _estimate_resolution_limit(
         m.set_limit_cc_half(params.cc_half)
         m.set_cc_half_fit(params.cc_half_fit)
         m.set_cc_half_significance_level(params.cc_half_significance_level)
-        if use_isigma:
+        if params.isigma:
```
rjgildea: There is no need for this if statement here as setting the limit isn't what determines what is actually used for the cutoff.

Modules/Scaler/CommonScaler.py
```diff
@@ -990,9 +984,9 @@ def _estimate_resolution_limit(
...
             m.set_limit_isigma(params.isigma)
-        if use_misigma:
+        if params.misigma:
```
rjgildea: As above, if statement not needed.

Handlers/Phil.py
```diff
@@ -939,6 +939,22 @@
...
+  resolution {
+    isigma = None
+    misigma = None
+  }
+}
+"""
+    )
+)
+
+master_phil = phil_scope.fetch(sources=[phil_overrides])
```
rjgildea: I don't think it is necessary to assign to an intermediate `phil_overrides` variable here, you can directly assign to `master_phil` above at line 942 (and no need to rename the variable at line 17).

jbeilstenedmands:
Thanks for the suggestions, now adopted. I think we're in agreement about the code changes needed. Guess we need a consensus now on whether to go ahead with the behaviour change.

PR #380, opened 2019-11-19
Remove some more dead code

Anthchirp:
I used this heuristic:
* `grep` source code for `def .*(`
* for every name obtained thusly
  * grep the entire code base for that name
  * if the name only appears once remove the definition
* `goto 10` until no more instances found

This PR is the outcome (with 11 false positives discounted)


Wrappers/Dials/Symmetry.py
```diff
@@ -71,14 +71,6 @@ def set_mode_absences_only(self):
             self._laue_group = None
             self._sys_abs_check = True
 
-        def set_mode_laue_only(self):
```
jbeilstenedmands: Although these are not currently used, probably useful to keep around for future use

Anthchirp: If these are really needed in the future wouldn't it make more sense to add them back in as properties instead of getter/setter methods?

PR #406, opened 2020-01-20
Multiplex: add radar plot for comparison of merging statistics

rjgildea:
Add a [radar plot ](https://en.wikipedia.org/wiki/Radar_chart) to xia2.multiplex html report for comparison of merging statistics between clusters. Each attribute is scaled between 0->1 so that the "best" datasets for each attribute is equal to 1. Currently plotted are:
- CC½
- Multiplicity
- Completeness
- I/σ(I)
- Resolution (1/d^2)
Example output from:
```
$ xia2.multiplex $(dials.data get -q multi_crystal_proteinase_k)/*.{json,pickle} max_clusters=2
```
Screenshot 2020-01-20 at 13 14 19


ndevenish:
Any reason not to pull this in?

PR #407, opened 2020-01-23
Rename xia2.error to xia2-error.txt

rjgildea:
It's a text file, so call it .txt. This will facilitate viewing the contents of the file in phone/web email clients.


Anthchirp:
(note that this will require changes in dlstbx, too)

dagewa:
I will need to remember to update the xia2/dials ccp4i2 interface when this gets to CCP4, as it looks for the error file.

Issue #408, opened 2020-01-24
xia2 running on a directory - check if .nxs aliases .h5 master data


graeme-winter:
SCI-8781 - 

Can have .nxs aliasing same data as master.h5 - if directory passed same pixels will be processed twice. Don't do this.

graeme-winter:
Example - current behaviour gives

```
BEGIN PROJECT AUTOMATIC
BEGIN CRYSTAL DEFAULT

BEGIN WAVELENGTH NATIVE
WAVELENGTH 0.980274
END WAVELENGTH NATIVE

BEGIN SWEEP SWEEP1
WAVELENGTH NATIVE
DIRECTORY /Users/graeme/data/i04-eiger-small
IMAGE Therm_6_2.nxs
START_END 1 488
END SWEEP SWEEP1

BEGIN SWEEP SWEEP2
WAVELENGTH NATIVE
DIRECTORY /Users/graeme/data/i04-eiger-small
IMAGE Therm_6_2_master.h5
START_END 1 488
END SWEEP SWEEP2

END CRYSTAL DEFAULT
END PROJECT AUTOMATIC
```

graeme-winter:
First order fix: avoid data sets which are precisely aliased, so key on `data_file:start:end`

PR #409, opened 2020-01-24
408 alias

graeme-winter:
Fix the aliasing issue identified in #408 

The fix _is_ a spot convoluted but I think this recognises the truth of the situation


graeme-winter:
Two of the commits here are cherry picks so suggest don't squash merge.

graeme-winter:
Merging based on +ve feedback from @Anthchirp

PR #414, opened 2020-02-13
refactor frenchwilson calling code

Anthchirp:
Turn the object with a single method writing to stdout and taking a 3-element PHIL object
into a normal function with 3 parameters returning a string.

Change the wrapper to then not run this function in a subprocess.

Regression testing shows a tolerance warning, but I think that was already there.


Anthchirp:
re-ran all tests and my conclusion is that this strictly improves things

PR #415, opened 2020-02-14
Refactor xia2.overload

graeme-winter:
Refactor to use Phil parameters, OptionParser, add a little test that this actually runs, using centroid_test_data

Making a draft PR at the moment so we can continue to improve the structure before merging back.


ndevenish:
As Feb 2020, I doubt this is immediately being worked on. I've merged and fixup'd the existing changes a little, so that we can merge what is already changed.

PR #416, opened 2020-02-14
Be more explicit when setting model options for dials.scale

jbeilstenedmands:
Currently, the dials scaling model options are not being set very cleanly. This seems to be causing an issue for xia2.multiplex - where for narrow wedge scans < 15 degrees, only two parameters are used for scale and decay components, but an absorption surface is used (24 params), which is the opposite default behaviour to dials.scale. There were also lots of Phil options which probably werent working as expected.
E.g. This is the current model parameterisation (+ absorption surface)

newplot

This PR updates things to follow more closely the default dials behaviour while preserving the current xia2 Phil options. This restores behaviour expected for dials.scale, giving the following model parameterisation (no absorption surface), which I assert is generally more useful for multi-crystal data:

newplot (2)

### DIFF
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -401,10 +401,10 @@ dials
     .expert_level = 1
     .short_caption = "Scaling"
   {
-    model = *auto physical array kb
+    model = *auto physical array KB
       .type = choice
       .help = "Choice of scaling model parameterisation to apply"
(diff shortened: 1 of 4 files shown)


jbeilstenedmands:
Any objections to me merging these changes? It's holding up some other work on scaling models I want to push into dials.

PR #420, opened 2020-02-20
Reindex to input spacegroup, not pointgroup

### LINKED ISSUE #419: xia2 pipeline=dials space_group= gives incorrect space group in output mtz
```
module load dials/latest
image_file=`ls $(dials.data get -q x4wide)/*0001.cbf`
xia2 image=$image_file space_group=P41212
```
The correct space group is reported in the log file:
```
$ grep Assuming xia2.txt 
Assuming spacegroup: P 41 21 2
```
However, the output mtz file is in the Laue group, not the space group:
```
$ iotbx.mtz.dump DataFiles/AUTOMATIC_DEFAULT_scaled_unmerged.mtz | grep Space
Space group symbol from file: P422
Space group number from file: 89
Space group from matrices: P 4 2 2 (No. 89)
```

rjgildea:
If the space group was set, xia2 would report the correct space group in the log file, but the mtz file would be in the Laue group.

Exercise space_group= option in various pipelines, check that the output data files contain the expected space group.

Fixes #419

### DIFF
--- a/Modules/Scaler/DialsScaler.py
+++ b/Modules/Scaler/DialsScaler.py
@@ -217,15 +217,14 @@ class DialsScaler(Scaler):
         return need_to_return
 
     def _input_pointgroup_scale_prepare(self):
-        # is this function completely pointless?
-        # ---------- REINDEX ALL DATA TO CORRECT POINTGROUP ----------
-        ####Redoing batches only seems to be in multi_sweep_idxing for CCP4A
+        pointgroup = self._scalr_input_pointgroup
         if self._scalr_input_spacegroup:
             self._scalr_likely_spacegroups = [self._scalr_input_spacegroup]
-        logger.debug("Using input pointgroup: %s", self._scalr_input_pointgroup)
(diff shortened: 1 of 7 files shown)


Anthchirp:
how much does this cost in terms of time added for a full regression run?

rjgildea:
> how much does this cost in terms of time added for a full regression run?

This is why I changed the test to do quick processing on just 10 images. IMO it is worth the added time in order to have proper test coverage of a widely used option (indeed, an option we routinely use on the beamlines).

```
$ time pytest Test/regression/test_X4_wide.py::test_space_group --regression-full -vs
Results (1386.30s):
       6 passed

real	23m8.409s
user	8m30.491s
sys	1m4.080s
```

rjgildea:
> I am wondering though why we have `spacegroup` and `space_group` in the source code 🙂

"Historical reasons". `spacegroup` comes from original xia2 code, `space_group` comes from cctbx naming.

PR #424, opened 2020-02-28
Remove the Environment singleton

Anthchirp:
Attach a path to the XProject object instead of keeping a global variable that denotes the initial working directory.
This should solve most of the occurrences where files appear in the source directory when running tests.


Anthchirp:
The regression tests pass, however three tests need some remedial work

Anthchirp:
All regression tests now pass

PR #426, opened 2020-03-05
Add fix_distance parameter

dagewa:
This provides an easy way to fix the detector distance for indexing and refinement. This can be useful for low resolution cases, or electron diffraction, where detector distance refinement may be unstable. This provides more robust refinement without being as severe as the full `fix_geometry`

### DIFF
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -207,6 +207,11 @@ dials
             "Most useful when also providing a reference geometry to xia2."
     .short_caption = "Fix geometry"
     .expert_level = 1
+  fix_distance = False
+    .type = bool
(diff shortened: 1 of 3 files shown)


Handlers/Phil.py
```diff
@@ -207,6 +207,11 @@
             "Most useful when also providing a reference geometry to xia2."
     .short_caption = "Fix geometry"
     .expert_level = 1
+  fix_distance = False
+    .type = bool
+    .help = "Whether or not to refine the detector distance in dials.index and dials.refine. " \
```
Anthchirp:
```suggestion
    .help = "Do not refine the detector distance in dials.index and dials.refine."
```
gives clearer indication what `True` and `False` mean. Assume the trailing `\` is not relevant

PR #428, opened 2020-03-05
Add an auto option for full_matrix minimisation in Dials scaling.

jbeilstenedmands:
This would run full matrix scaling only if 4 sweeps or fewer. Perhaps this is too low a number and 10 would be more appropriate? Or some check based on the number of reflections?

### DIFF
--- a/Handlers/Phil.py
+++ b/Handlers/Phil.py
@@ -439,10 +439,11 @@ dials
       .type = choice
       .help = "Choice of whether to refine an error model to adjust the" \
               "intensity sigmas using a two-parameter model."
-    full_matrix = True
+    full_matrix = auto
       .type = bool
-      .help = "Option to turn off GN/LM refinement round used to determine " \
-              "error estimates on scale factors."
+      .help = "Option to turn on/off Levenberg-Marquardt refinement round used" \
+              "to determine error estimates on scale factors. auto will set" \
+              "full_matrix=True if 4 sweeps or less."
     outlier_rejection = *standard simple
       .type = choice
       .help = "Choice of outlier rejection routine. Standard may take a " \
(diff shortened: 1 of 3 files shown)


graeme-winter:
Thanks @jbeilstenedmands for the quick work with this - perhaps the best solution is to work out the point where the full-matrix becomes time punishing (reflections / parameters / sweeps / ...) and set the switch point somewhere below this - though I appreciate that this is not an easy question to answer. 

One interesting point would be to assess the impact on the data coming out - if we have multiple small data sets as @dagewa pointed out the errors from the scaling models could be more significant hence a greater need for propagation. If however we have a number of large data sets the uncertainties on error model parameters are unlikely to significantly increase the uncertainties on the scaled intensities as the model uncertainties will be small - perhaps this is the metric to use?

graeme-winter:
I do wonder (though I recognise a feature creep when I see one) whether this is the time to set "auto" for absorption correction etc.? Thinking again of the large N small wedges case there is not really enough information to do a proper harmonic correction - unless this is handled by `model=auto` and it defaulting to kB?

graeme-winter:
Example case here of 50 sweeps of a small molecule data set dropped from 15 hours to 6 minutes by switching off the full matrix minimisation :roll_eyes:

dagewa:
> perhaps the best solution is to work out the point where the full-matrix becomes time punishing

Ideally we would find a point that satisfies both this requirement and that does not show significant difference in results between full matrix on or off. It would be nice to avoid a discontinuity in results at the switchover point, if possible.

dagewa:
Or is it the case that the results are never that different, because full matrix is just done for one cycle to get error estimates?

rjgildea:
I wonder if this might be better implemented as a `full_matrix=auto` option in dials.scale itself?

1. dials.scale users would also have the benefit of any `auto` option
2. dials.scale probably has more information at this point about the nature of the specific scaling/minimisation problem and the relative cost/benefit of including a full matrix cycle.

PR #432, opened 2020-03-28
Use recalculated_unit_cell in two_theta_refine

rjgildea:
Instead pass dials.two_theta_refine output experiments directly to dials.scale.
Ensure that the dials.two_theta_refine cell is used in downstream output files.

See cctbx/dxtbx#153 and dials/dials#1214

### DIFF
--- a/Test/regression/test_X4_wide.py
+++ b/Test/regression/test_X4_wide.py
@@ -103,7 +103,7 @@ def test_dials(regression_test, dials_data, tmpdir, ccp4):
         dials_data("x4wide").strpath,
     ]
     result = procrunner.run(command_line, working_directory=tmpdir)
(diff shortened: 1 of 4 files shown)


Test/regression/test_X4_wide.py
```diff
@@ -101,7 +103,24 @@ def test_dials(regression_test, dials_data, tmpdir, ccp4):
         dials_data("x4wide").strpath,
     ]
     result = procrunner.run(command_line, working_directory=tmpdir)
-    print(result)
+    scaled_expt_file = tmpdir.join("DataFiles/AUTOMATIC_DEFAULT_scaled.expt")
```
Anthchirp, line 106:
```suggestion
    scaled_expt_file = tmpdir / "DataFiles" / "AUTOMATIC_DEFAULT_scaled.expt"
```

PR #433, opened 2020-03-30
Extend xia2.to_shelxcde to support MAD

isikhar:


graeme-winter:
<- reviewing

graeme-winter:
Supports MAD iff you provide data in the sequence peak, infl, hrem, lrem - case I tested here has infl and lrem (to maximise dispersive difference, from 1e3 years ago @ JCSG) but I don't see how I can pass this information in

Propose:

`xia2.to_shelxcde peak=? infl=? lrem=? hrem=? sad=?`

As a way of breaking this ambiguity - opinions?

graeme-winter:
Also, if we have the case where there is a merged MTZ file with the correct wavelength names assigned, would we like to be able to use this? For example where one MTZ file has 2 or more wavelengths in as attached...

[433-example.zip](https://github.com/xia2/xia2/files/4409635/433-example.zip)

rjgildea:
I notice that there aren't any tests either for the existing behaviour of xia2.to_shelxcde, or for the new behaviour added in this PR - now might be a good opportunity to increase our test coverage. There are _some_  processed data files in dials_data that could potentially be used for writing a test. Perhaps not for the MAD case, but these could be easily added to dials_data. Alternatively we could tack a test onto the end of a full xia2 test, but a standalone test would probably be preferable. Let me know if you need a hand with adding a test 🙂

command_line/to_shelxcde.py
```diff
@@ -1,54 +1,179 @@
...
+                      set(('sad', 'label', 'nat', 'prefix', 'sites')),
+                      set(('mad', 'peak', 'infl', 'hrem', 'lrem', 'nat', 'prefix', 'sites')))
+    if not any(param_names.issubset(grp) for grp in allowed_groups):
+        parser.error('Invalid combination of the input parameters')
+
+    if 'sad' in params:
+        shelx_files = export_sad_hkl(params)
+    elif 'mad' in params:
+        shelx_files = export_single_mad_hkl(params)
+    elif 'peak' in params or 'infl' in params or 'hrem' in params or 'lrem' in params:
```
Anthchirp, line 168:
```suggestion
    elif any(p in params for p in ("peak", "infl", "hrem", "lrem")):
```

command_line/to_shelxcde.py
```diff
@@ -1,54 +1,179 @@
...
+        shelx_files = export_mad_hkl(params)
     else:
-        to_shelxcde(sys.argv[1], sys.argv[2])
+        raise ValueError("Input dataset file not specified")
```
Anthchirp, line 171: since the function builds on `argparse` anyway this is probably more helpful to the user:
```suggestion
        parser.error("Input dataset file not specified")
```

(untested)

graeme-winter:
MTZ file in this ticket _could_ be used to create a test case for this with 2 wavelength MAD (infl, lrem) however not making this a precondition for merging. 

```
xia2.to_shelxcde --sites 7  --infl ../xia2-12287-mad-datafiles/AUTOMATIC_DEFAULT_scaled_unmerged_WAVE1.mtz --lrem ../xia2-12287-mad-datafiles/AUTOMATIC_DEFAULT_scaled_unmerged_WAVE2.mtz 2isb
bash 2isb.sh 
shelxd 2isb_fa
shelxe 2isb 2isb_fa -h 
```

seemed to do sensible things, solved structure so 👍

PR #439, opened 2020-04-14
Make Handlers.Files.cleanup() a contextmanager

rjgildea:
Ensure that cleanup() is called even when an exception occurs during
processing. This also means recorded data/log files will always be
copied to DataFile/LogFiles directories respectively, even when an
exception has occurred.

### DIFF
--- a/Handlers/Files.py
+++ b/Handlers/Files.py
@@ -20,14 +20,9 @@ class _FileHandler(object):
         self._temporary_files = []
         self._output_files = []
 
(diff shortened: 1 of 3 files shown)


graeme-winter:
100% behind the idea, I think it is a very good one. 

Not a request: I wonder if it would be good to refactor xia2_main _slightly_ - the additional indentation shows that there are already too many layers of indentation in there 🙃 - welcome opinions. Will test in a mo

PR #441, opened 2020-04-28
Float some 80% of all imports

Anthchirp:
This is a subset of all float-able imports, courtesy of @ndevenish's script.
We can't float all the imports as there are some circular dependencies.
I've unfloated those files that take part in the loops.
I then ran `isort` over most of the remaining files.

Running tests now.


graeme-winter:
Scanning the change set this does touch a fair bit of rather rustic code... 🤔 every time I see stuff like this the things-to-fix list gets appended to

Anthchirp:
tests passed (read: not more broken than before)

PR #449, opened 2020-05-12
Remove a redundant check for twinning

benjaminhwilliams:
Remove an old check for twinning, based on calculating the fourth moment of E.  Since xia2 now runs xtriage, this check is redundant and it also fails rather ungracefully if the number of reflections is small.


graeme-winter:
Reviewing the changes looks fine, but want to run tests before I properly review.

PR #452, opened 2020-05-14
Remove deprecated command-line options

graeme-winter:
Will probably be more to follow, but this retires the -3dii etc.


Anthchirp:
looking at removal of `-dials` etc. this could probably do with a `newsfragments/452.removal`?
Should we also increment version 0.6 -> 0.7?

graeme-winter:
> looking at removal of `-dials` etc. this could probably do with a `newsfragments/452.removal`?
> Should we also increment version 0.6 -> 0.7?

Good suggestions both, implemented

PR #461, opened 2020-05-20
Remove xia2.chef: broken and now replaced by dials.damage_analysis

### LINKED ISSUE #460: xia2.chef broken by recent DIALS changes
https://github.com/dials/dials/commit/127d20ceec7cc262ef3dc5982147cf632a041f60 broke `xia2.chef`.  Reported by I19 staff at DLS.  Do we want to retire `xia2.chef` in favour of `dials.damage_analysis`?  I note that the latter is not yet documented on the DIALS website.

rjgildea:
Cleanup other unused chef-related dead code.

Fixes #460


benjaminhwilliams:
Would it be worth replacing `xia2.chef` with a simple script that prints a deprecation warning pointing users to `dials.damage_analysis`, for one release cycle?

rjgildea:
> Would it be worth replacing `xia2.chef` with a simple script that prints a deprecation warning pointing users to `dials.damage_analysis`, for one release cycle?

I'm not sure this is really necessary - the command will no longer exist which should alert the user that something has changed. And the release notes will contain the information about its replacement.

PR #466, opened 2020-05-29
Re-estimate resolution limit after deltacchalf filtering

rjgildea:
Sometimes the resolution of the data can improve significantly after deltacchalf filtering. Therefore if some experiments have been removed by deltacchalf filtering, perform full re-scaling of the filtered data in order to re-estimate the resolution limit.

Depends on dials/dials#1276


rjgildea:
One example where the resolution improves significantly from 2.67 Å to 2.26 Å after filtering. Without this PR, the filtered data set would be truncated to 2.67 Å.
image
image
image

PR #468, opened 2020-06-01
Remove xia2-files.txt

graeme-winter:
Instead send the corresponding text to xia2-debug.txt via logger.debug()


Anthchirp:
needs a few changes, commit forthcoming

PR #476, opened 2020-06-29
If nimages = 1 and ntrigger > 1 treat as 1 sweep

### LINKED ISSUE #475: image=foo.h5 multiplied by #frames in foo.h5
Particularly visible as #frames == 9000 in this case

Datum: data from NSLS II

`dials.import` does something sensible => comes out with 9,000 image sweep

graeme-winter:
Fixes #475

If ntrigger > 1 but 1 image per trigger, treat this as single scan of ntrigger images


graeme-winter:
Thanks to @Anthchirp for comment on readability of news

PR #485, opened 2020-07-08
Accept multiple reference instrument models in a single file

benjaminhwilliams:
At present, xia2 expects that each `reference_geometry=` argument passed at the command line consists of a experiment list `.expt` file containing just one experiment.  If the file contains multiple experiments, two undesirable things happen:
* The experiment list is loaded from file with `check_format=True`, which leaves xia2 exposed to `FileNotFoundError`s when reference instrument models contain invalid references to hot mask files and/or image templates.
* If the experiment list is loaded successfully, only the first experiment is used and the others are discarded.

Here, I introduce a new workflow:
1. Always load reference instrument model experiment lists with `check_format=False`, since we have no interest in creating an imageset and don't want to be troubled by broken file references.
2. Aggregate all the experiments from experiment lists passed as instrument models with `reference_geometry=`.
3. Remove equivalent instrument models to produce a minimal set of unique reference instrument models.

### DIFF
--- a/Schema/__init__.py
+++ b/Schema/__init__.py
@@ -260,14 +260,15 @@ def load_reference_geometries(geometry_file_list):
             # as per dials.show, rather than the UID string of the experiment.
             logger.debug(f"Experiment {b[3]} of {b[2]} is a duplicate.")
         else:
-            unique.update((a, b))
+            unique.add(a)
+            unique.add(b)
 
     n = len(unique)
     logger.debug(f"Found {n} unique reference geometr{'ies' if n != 1 else 'y'}.")
     for geometry in unique:
         logger.debug(f"Experiment {geometry[3]} of {geometry[2]} is unique.")
 
-    return [dict(zip(("detector", "beam"), components[:2])) for components in unique]
+    return [{"detector": components[0], "beam": components[1]} for components in unique]
 
 
 def compare_geometries(detectorA, detectorB):
--- a/Schema/__init__.py
+++ b/Schema/__init__.py
@@ -4,10 +4,9 @@ import itertools
 import logging
 import os
 
+from dxtbx.model import ExperimentList
 from scitbx.array_family import flex
-
 from xia2.Handlers.Phil import PhilIndex
-
 
 logger = logging.getLogger("xia2.Schema")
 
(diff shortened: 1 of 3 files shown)


Schema/__init__.py
```diff
@@ -246,34 +245,29 @@ def update_with_reference_geometry(imagesets, reference_geometry_list):
...
+
+    logger.debug("Removing duplicate reference geometries.")
+    unique = set()
+    for a, b in filter(unique.isdisjoint, itertools.combinations(ref_components, 2)):
+        if compare_geometries(a[0], b[0]):
+            # Note that expt_index is the index of the experiment in the expt list file,
+            # as per dials.show, rather than the UID string of the experiment.
+            logger.debug(f"Experiment {b[3]} of {b[2]} is a duplicate.")
+        else:
+            unique.update((a, b))
```
Anthchirp, line 263: Suggestion:
```suggestion
            unique.add(a)
            unique.add(b)
```
does the same, but feels more obvious to me. I can't remember seeing `.update()` on a `set` before

Schema/__init__.py
```diff
@@ -246,34 +245,29 @@ def update_with_reference_geometry(imagesets, reference_geometry_list):
...
-                combination[1]["file"],
-            )
-            raise Exception("Reference geometries too similar")
-    return reference_components
+    n = len(unique)
+    logger.debug(f"Found {n} unique reference geometr{'ies' if n != 1 else 'y'}.")
+    for geometry in unique:
+        logger.debug(f"Experiment {geometry[3]} of {geometry[2]} is unique.")
+
+    return [dict(zip(("detector", "beam"), components[:2])) for components in unique]
```
Anthchirp, line 270: again this feels unnecessary obscure.
I suspect it's equivalent to this?
```suggestion
    return [{"detector": components[0], "beam": components[1]} for components in unique]
```

Test/Schema/test_Schema.py
```diff
@@ -18,3 +19,22 @@ def test_load_imageset(dials_data, tmp_path):
...
+
+    Test the function that finds the set of unique instrument models from a list
+    of experiment list files.
+    """
+    files = ["scaled_20_25.expt", "scaled_30.expt", "scaled_35.expt"]
+    files = [(dials_data("l_cysteine_4_sweeps_scaled") / f).strpath for f in files]
+
+    # Check that there are four input instrument models, of which only two are unique.
+    assert (
+        sum([len(ExperimentList.from_file(f, check_format=False)) for f in files]) == 4
```
Anthchirp, line 36:
```suggestion
        sum(len(ExperimentList.from_file(f, check_format=False)) for f in files) == 4
```
no explicit list generation required where iterators can be used

benjaminhwilliams: Yep, quite right.  Thanks.

Schema/__init__.py
benjaminhwilliams: Good point.  I think I'd gone around the houses a few times!

Schema/__init__.py
benjaminhwilliams: Sure thing.

benjaminhwilliams:
Thanks for the review @Anthchirp, I'll squash & merge when the CI checks go green.

PR #491, opened 2020-07-15
Pass image_range to ExperimentListTemplateImporter

### LINKED ISSUE #490: xia2 complains about missing images even if outside of specified range
```
$ cp $(dials.data get -q x4wide)/*.cbf .
$ xia2 image=X4_wide_M1S4_2_0001.cbf:1:29 read_all_image_headers=false
```
Related to (but not entirely solved by) dials/dials#1333 and cctbx/dxtbx#191.

rjgildea:
This means that xia2 doesn't complain about missing images if they lie outside the start:end specified on the command line.
Only affects read_image_headers=False (i.e. loading the dxtbx imageset from a template).

Fixes #490.

See also cctbx/dxtbx#191 and dials/dials#1333.

### DIFF
--- a/Schema/__init__.py
+++ b/Schema/__init__.py
@@ -7,7 +7,6 @@ import os
 from dxtbx.model import ExperimentList
 from dxtbx.model.experiment_list import ExperimentListTemplateImporter
 from scitbx.array_family import flex
-from xia2.Handlers.CommandLine import CommandLine
 from xia2.Handlers.Phil import PhilIndex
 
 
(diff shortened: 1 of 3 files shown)


rjgildea:
Needs a test adding.

rjgildea:
If start:end is not set for a given template, then the behaviour should be exactly as before this PR. Standalone testing of this area of the code is made tricky by the excessive use of singletons (e.g. CommandLine object) that require jumping through some hoops to reliably test outside of running a full xia2 job (which is expensive).

graeme-winter:
Please could you remerge master into this so we have the resolutionizer fixes? Without this cannot test...

graeme-winter:
<- would like to test e.g. `xia2 image=~/data/foo/bar_0001.cbf.gz:1:2400:800`

graeme-winter:
@rjgildea  👍 

Testing now but 1st glance seems to be all good

PR #499, opened 2020-07-23
Fix load_imagesets introduced by #491

### LINKED ISSUE #498: Error running multiple image ranges same template with read_all_image_headers=False
New bug introduced by #491:
```
$ xia2 image=$(dials.data get -q x4wide)/X4_wide_M1S4_2_0001.cbf:1:45 image=$(dials.data get -q x4wide)/X4_wide_M1S4_2_0001.cbf:46:90 read_all_image_headers=False
Environment configuration...
Python => /Users/rjgildea/software/cctbx_py3/build/../conda_base/python.app/Contents/MacOS/python
CCTBX => /Users/rjgildea/software/cctbx/modules
CCP4 => /Applications/ccp4-7.1
CCP4_SCR => /var/folders/dr/_bgcxh092cg7pt001tzwvl5c0000gn/T/tmpjgzy0n7q
Starting directory: /Users/rjgildea/tmp/012
Working directory: /Users/rjgildea/tmp/012
Free space:        23.79 GB
Host: Richards-MBP-10
XIA2 0.7.69-g2e1d23b5
DIALS 3.dev.138-ge95e02e1a
CCP4 7.1.000
Command line: xia2 image=/Users/rjgildea/software/cctbx_py3/build/dials_data/x4wide/X4_wide_M1S4_2_0001.cbf:1:45 image=/Users/rjgildea/software/cctbx_py3/build/dials_data/x4wide/X4_wide_M1S4_2_0001.cbf:46:90 read_all_image_headers=False
------------------------------------------
| No images assigned for crystal DEFAULT |
------------------------------------------
Project directory: /Users/rjgildea/tmp/012
Error: no Integrater implementations assigned for scaling
xia2.support@gmail.com
```

The problem is that `load_imagesets` now returns multiple imagesets covering each individual image range, rather than a single imageset covering the whole image range as it did previously:
https://github.com/xia2/xia2/blob/f7a667aa0816253f14b7898b7580072b3664279a/Schema/__init__.py#L151-L163

This ultimately ends up failing here:
https://github.com/xia2/xia2/blob/f7a667aa0816253f14b7898b7580072b3664279a/Applications/xia2setup.py#L454-L459

rjgildea:
Don't overwrite input kwarg in `load_imagesets`:

This meant that `load_imagesets` only returned one of the imagesets in cases that it should have returned multiple imagesets with the same template.

Also include subsequent bug fix for sorting sweeps with matching epochs, i.e. two sweeps with the same template will have the same epoch, resulting in a `TypeError: '<' not supported` error.

Fixes #498

### DIFF
--- a/Applications/xia2setup.py
+++ b/Applications/xia2setup.py
@@ -306,9 +306,14 @@ def _write_sweeps(sweeps, out):
     for sweep in sweeplist:
         sweeps = _known_sweeps[sweep]
 
-        # sort on exposure epoch
-        epochs = [s.get_imageset().get_scan().get_epochs()[0] for s in sweeps]
-        sweeps = [s for _, s in sorted(zip(epochs, sweeps))]
+        # sort on exposure epoch followed by first image number
+        sweeps = sorted(
+            sweeps,
+            key=lambda s: (
+                s.get_imageset().get_scan().get_epochs()[0],
+                s.get_images()[0],
+            ),
+        )
         for s in sweeps:
             if len(s.get_images()) < min_images:
                 logger.debug("Rejecting sweep %s:" % s.get_template())
@@ -424,9 +429,14 @@ def _write_sweeps(sweeps, out):
     for sweep in sweeplist:
         sweeps = _known_sweeps[sweep]
 
-        # sort on exposure epoch
-        epochs = [s.get_imageset().get_scan().get_epochs()[0] for s in sweeps]
-        sweeps = [s for _, s in sorted(zip(epochs, sweeps))]
+        # sort on exposure epoch followed by first image number
+        sweeps = sorted(
+            sweeps,
+            key=lambda s: (
+                s.get_imageset().get_scan().get_epochs()[0],
+                s.get_images()[0],
+            ),
+        )
 
         for s in sweeps:
             # require at least n images to represent a sweep...
--- a/Schema/__init__.py
+++ b/Schema/__init__.py
@@ -154,11 +154,11 @@ def load_imagesets(
                 start_ends = CommandLine.get_start_ends(full_template_path)
                 if not start_ends:
                     start_ends.append(None)
-                for image_range in start_ends:
+                for start_end in start_ends:
                     importer = ExperimentListTemplateImporter(
                         [full_template_path],
                         format_kwargs=format_kwargs,
-                        image_range=image_range,
+                        image_range=start_end,
                     )
                     experiments.extend(importer.experiments)
 
--- a/Test/Applications/test_xia2setup.py
+++ b/Test/Applications/test_xia2setup.py
@@ -21,10 +21,14 @@ def test_write_xinfo_split_sweep(dials_data, tmp_path, monkeypatch):
             "read_all_image_headers=False",
         ],
     ):
-        # Force reload of CommandLine singleton object with the parameters above
+        # Force reload of various singleton objects with the parameters above
         import xia2.Handlers.CommandLine
+        import xia2.Applications.xia2setup
+        import xia2.Schema
 
         reload(xia2.Handlers.CommandLine)
+        reload(xia2.Applications.xia2setup)
+        xia2.Schema.imageset_cache = xia2.Schema._ImagesetCache()
 
         template = xia2.Handlers.CommandLine.CommandLine.get_template()
         directories = [os.path.split(p) for p in template]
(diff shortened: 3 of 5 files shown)


rjgildea:
> Bit concerned about the heap of new technical debt for the CommandLine singleton

I'm not entirely happy about it, but I don't see an obvioius way to avoid it without running a full xia2 job via procrunner.

graeme-winter:
This fixes the bug, and improves the test coverage, so 0th order this is a good change set. Full review incoming.

Applications/xia2setup.py
```diff
@@ -306,9 +306,14 @@ def _write_sweeps(sweeps, out):
     for sweep in sweeplist:
         sweeps = _known_sweeps[sweep]
 
-        # sort on exposure epoch
-        epochs = [s.get_imageset().get_scan().get_epochs()[0] for s in sweeps]
-        sweeps = [s for _, s in sorted(zip(epochs, sweeps))]
+        # sort on exposure epoch followed by first image number
```
graeme-winter, line 309: 👍  less weird way of doing the sorting

Applications/xia2setup.py
```diff
@@ -424,9 +429,14 @@ def _write_sweeps(sweeps, out):
...
+        sweeps = sorted(
```
graeme-winter, line 433: duplicate: wonder if `def sort_sweeps()` would help us here?

Schema/__init__.py
```diff
@@ -154,11 +154,11 @@ def load_imagesets(
                 start_ends = CommandLine.get_start_ends(full_template_path)
                 if not start_ends:
                     start_ends.append(None)
-                for image_range in start_ends:
+                for start_end in start_ends:
```
graeme-winter, line 157: 👍

Test/Applications/test_xia2setup.py
```diff
@@ -0,0 +1,39 @@
+import mock
+import os
+import sys
+from importlib import reload
+
+from xia2.Handlers.XInfo import XInfo
+
+
+def test_write_xinfo_split_sweep(dials_data, tmp_path):
```
graeme-winter, line 9: Could consider writing this test to execute `xia2setup` via procrunner, except for the fact that that is now gone - yes this is probably a reasonable approach to executing the tests. Can extend this as well to cover other use cases described in the issue this fixes.

graeme-winter:
@rjgildea I tried adding a second test as

```
def test_write_xinfo_unroll(dials_data, tmp_path):
    # This test partially exercises the fix to https://github.com/xia2/xia2/issues/498 with a different syntax
    xinfo = tmp_path.joinpath("unroll.xinfo")

    with mock.patch.object(
        sys,
        "argv",
        [
            f"image={dials_data('insulin').join('insulin_1_001.img:1:45:15')}",
            "read_all_image_headers=False",
        ],
    ):
```

but this fails... which makes me wonder if this testing scheme is fragile?

graeme-winter:
Tried adding a shim to `CommandLine`
```
def _reload():
    """Reload singleton, for use in testing environments only."""
    global CommandLine
    CommandLine = _CommandLine()
    CommandLine.setup()
```

This does not help... 🤔

graeme-winter:
> sorting looks fine.
> 
> Bit concerned about the heap of new technical debt for the CommandLine singleton

I think the CommandLine singleton (and I suspect other related ones...) are already a heap of technical debt... as described above there are still gaps even with reloading it

Not asking for this to be fixed in this PR, but we should look to see if it _can_ be fixed sometime

graeme-winter:
@rjgildea quick pause on this: the test _does_ leave droppings in the form of a brace of PHIL files

graeme-winter:
> @rjgildea quick pause on this: the test _does_ leave droppings in the form of a brace of PHIL files

Got a fix; incoming

Test/Applications/test_xia2setup.py
```diff
@@ -0,0 +1,41 @@
...
+import sys
+from importlib import reload
+
+from xia2.Handlers.XInfo import XInfo
+
+
+def test_write_xinfo_split_sweep(dials_data, tmp_path, monkeypatch):
+    # This test partially exercises the fix to xia2/xia2#498
+    # run in tmp dir so droppings not cluttering source area
+    monkeypatch.chdir(tmp_path)
```
rjgildea, line 12: Why not use the `run_in_tmpdir` fixture that we use everywhere else?

graeme-winter: For some reason I thought that was about running the things you run in a tmpdir, rather than running the code in there, and google gave me this monkeypatch solution which worked 😉

Respectfully acknowledge that a more standard solution is better 👍

rjgildea:
> @rjgildea I tried adding a second test as
> 
> ```
> def test_write_xinfo_unroll(dials_data, tmp_path):
>     # This test partially exercises the fix to https://github.com/xia2/xia2/issues/498 with a different syntax
>     xinfo = tmp_path.joinpath("unroll.xinfo")
> 
>     with mock.patch.object(
>         sys,
>         "argv",
>         [
>             f"image={dials_data('insulin').join('insulin_1_001.img:1:45:15')}",
>             "read_all_image_headers=False",
>         ],
>     ):
> ```
> 
> but this fails... which makes me wonder if this testing scheme is fragile?

I have spent the past hour trying to figure out a way around this, but there are just too many singletons all over the place to make this work cleanly.

The first is `xia2.Schema.imageset_cache`, which means that when a second test is added, it actually retains the set of imagesets loaded in the first test.

Second, reloading the `CommandLine` module doesn't actually clean up all instances of the `CommandLine` object that have been previously imported inside other modules that have already been imported. So in some places `CommandLine.get_start_ends()` returns the correct values for this test, and in others it returns the values for the previously run test.

graeme-winter:
> > @rjgildea I tried adding a second test as
> > ```
> > def test_write_xinfo_unroll(dials_data, tmp_path):
> >     # This test partially exercises the fix to https://github.com/xia2/xia2/issues/498 with a different syntax
> >     xinfo = tmp_path.joinpath("unroll.xinfo")
> > 
> >     with mock.patch.object(
> >         sys,
> >         "argv",
> >         [
> >             f"image={dials_data('insulin').join('insulin_1_001.img:1:45:15')}",
> >             "read_all_image_headers=False",
> >         ],
> >     ):
> > ```
> > 
> > 
> > but this fails... which makes me wonder if this testing scheme is fragile?
> 
> I have spent the past hour trying to figure out a way around this, but there are just too many singletons all over the place to make this work cleanly.
> 
> The first is `xia2.Schema.imageset_cache`, which means that when a second test is added, it actually retains the set of imagesets loaded in the first test.
> 
> Second, reloading the `CommandLine` module doesn't actually clean up all instances of the `CommandLine` object that have been previously imported inside other modules that have already been imported. So in some places `CommandLine.get_start_ends()` returns the correct values for this test, and in others it returns the values for the previously run test.

TL;DR - `xia2` is broken, at least from a testing perspective. I can respect this conclusion. 

Propose:
 - merge the PR as it does make the code better
 - reinstate some version of xia2.setup to run as a command line program
 - write tests against this

Test/Applications/test_xia2setup.py
rjgildea: The `run_in_tmpdir` fixture literally is just a wrapper around the `tmpdir` fixture that does a `tmpdir.chdir()` before running your test:
https://github.com/dials/dials/blob/9fbf9fcc2dafcb03a916cb7a83bd875e2584004c/conftest.py#L79-L84

Anthchirp: I don't like `run_in_tmpdir` but acknowledge it is useful because xia2 has so many implicit file operations :disappointed: 

The reason I don't like it is because any `chdir` operation changes the working directory, which is essentially a per-process singleton. I know we don't run any parallel threads here (`pytest -n` runs in processes), but still it feels unclean.

I made some headway towards what I believe is overall a better approach in #424 where I added a `.path` attribute (type pathlib.Path) to `XProject`, which now no longer depends on a working directory.

rjgildea:
Re-introduced the `xia2.setup` command line program to enable easier testing without having to resort to jumping through hoops to workaround various singleton instances. Merged in master so we now get the full azure testing - which now pass.

graeme-winter:
@rjgildea given that you have reintroduced `xia2.setup` and express confidence that everything is hunky-dory please feel free to squash & merge

PR #500, opened 2020-07-24
Run regression tests on Azure

Anthchirp:
Pronto.


Anthchirp:
Looking okay. Tests are failing because they are genuinely failing.

Also tests identified a Python 3.8 circular import issue in dials, but that is a separate thing.

PR #509, opened 2020-08-04
Useful error mesage if passed foo_000001.h5

### LINKED ISSUE #508: Catch xia2 image=foo_000001.h5 gracefully
Semi-frequent problem is people (very reasonably) pass the data file rather than the master file as input.

We should catch this gracefully. 

```
Grey-Area x :) $ xia2 image=~/data/i03-screen19/Protk_1/Protk_1_8_000001.h5 
Environment configuration...
Python => /Users/graeme/svn/cctbx/build/../conda_base/python.app/Contents/MacOS/python
CCTBX => /Users/graeme/svn/cctbx/modules
CCP4 => /Applications/ccp4-7.0
CCP4_SCR => /var/folders/hs/gvssfcd910s0jczjhjczj6h80000gn/T/tmpk5rxex96
Starting directory: /private/tmp/x
Working directory: /private/tmp/x
Free space:        128.05 GB
Host: Grey-Area
XIA2 0.7.78-g4b1f7549
DIALS 3.dev.149-g6a7d752e2
CCP4 7.0.076
Command line: xia2 image=/Users/graeme/data/i03-screen19/Protk_1/Protk_1_8_000001.h5
Error: no sweeps found
xia2.support@gmail.com
```

is not _helpful_ to our users

graeme-winter:
Check the input master files for:

 - presence of /entry
 - absence of /data

if any of the input files fail both tests then append to a bad file list, `sys.exit` if this list not empty with helpful message. Fixes #508.

### DIFF
--- a/Applications/xia2_main.py
+++ b/Applications/xia2_main.py
@@ -1,6 +1,6 @@
 # A top-level interface to the whole of xia2, for data processing & analysis.
 
-
+import glob
 import logging
 import math
 import os
@@ -78,7 +78,21 @@ def check_hdf5_master_files(master_files):
                 bad.append(filename)
 
     if bad:
-        sys.exit("Not master files: %s" % " ".join(bad))
+
+        dirs = set([os.path.split(b)[0] for b in bad])
+        masters = sum((glob.glob(os.path.join(d, "*_master.h5")) for d in dirs), [])
+        nxss = sum((glob.glob(os.path.join(d, "*.nxs")) for d in dirs), [])
+
+        message = (
+            "Provided input files not master files:\n  "
+            + "\n  ".join(bad)
+            + "\ndo you mean one of:\n  "
+            + "\n  ".join(masters)
+            + "\nor:\n  "
+            + "\n  ".join(nxss)
+        )
+
+        sys.exit(message)
 
 
 def get_command_line():
--- a/Applications/xia2_main.py
+++ b/Applications/xia2_main.py
@@ -78,8 +78,7 @@ def check_hdf5_master_files(master_files):
                 bad.append(filename)
 
     if bad:
-        filenames = " ".join(bad)
-        sys.exit(f"Not master files: {filenames}")
+        sys.exit("Not master files: %s" % " ".join(bad))
 
 
 def get_command_line():
--- a/Applications/xia2_main.py
+++ b/Applications/xia2_main.py
@@ -7,6 +7,8 @@ import os
 import platform
 import sys
 
+import h5py
+
 from dials.util import Sorry
 from xia2.Handlers.Citations import Citations
 from xia2.Handlers.Environment import df
@@ -63,6 +65,23 @@ def check_environment():
         )
 
 
+def check_hdf5_master_files(master_files):
+    """Check the input HDF5 master files look a little bit like HDF5 master
+    files and not just the data files: if the latter then sys.exit() with a
+    helpful message"""
+
+    bad = []
+
+    for filename in master_files:
+        with h5py.File(filename, "r") as f:
+            if b"/data" in f and b"/entry" not in f:
+                bad.append(filename)
+
+    if bad:
+        filenames = " ".join(bad)
+        sys.exit(f"Not master files: {filenames}")
+
+
 def get_command_line():
     from xia2.Handlers.CommandLine import CommandLine
 
@@ -93,6 +112,8 @@ def get_command_line():
 
         directories = [os.path.abspath(d) for d in directories]
         from xia2.Applications.xia2setup import write_xinfo
+
+        check_hdf5_master_files(CommandLine.get_hdf5_master_files())
 
         if CommandLine.get_template() or CommandLine.get_hdf5_master_files():
             write_xinfo(
--- a/Applications/xia2_main.py
+++ b/Applications/xia2_main.py
@@ -1,6 +1,7 @@
 # A top-level interface to the whole of xia2, for data processing & analysis.
 
 import glob
+import itertools
 import logging
 import math
 import os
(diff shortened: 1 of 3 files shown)


graeme-winter:
Question was raised: if you can tell that this is not a useful input file why not go look for the useful data?

- magic code is bad - the program should do only what it is told to
- user education - means people can learn what _should_ be passed
- they may have wanted us to only process the data in that h5 file e.g. 100/1000 frames
- it is possible also that there is more than one route to this data file => non-deterministic outcome

Also use of VDS in future would make a complete mess of this

Applications/xia2_main.py
```diff
@@ -63,6 +65,22 @@ def check_environment():
...
+
+    bad = []
+
+    for filename in master_files:
+        with h5py.File(filename, "r") as f:
+            if b"/data" in f and b"/entry" not in f:
+                bad.append(filename)
+
+    if bad:
+        sys.exit("Not master files: %s" % " ".join(bad))
```
rjgildea, line 81: I think this error message probably needs to be clearer still. Something along the lines of "You appear to have passed a HDF5 data file rather than master file as input." Maybe also including a hint for what we are expecting instead, i.e. `*_master.h5` or `*.nxs`
Would it be possible to at least look for related master files and point the user towards them (even if we don't automatically use the master file)?

Applications/xia2_main.py
```diff
@@ -63,6 +65,22 @@ def check_environment():
         )
 
 
+def check_hdf5_master_files(master_files):
```
rjgildea, line 68: Please could you add a test for this function? As a standalone function, without any dependencies on global singletons, it should be relatively straightforward to add a test for this - both correctly given `master.h5` as input, and given data `.h5` files as input.

newsfragments/509.feature
```diff
@@ -0,0 +1 @@
+HDF5 file input: if data_000001.h5 passed not master give useful error message.
```
rjgildea, line 1:
```suggestion
Report useful error message if given data_000001.h5 instead of master.h5 as input. 
```
Possible re-wording.

Applications/xia2_main.py
graeme-winter: @rjgildea like this? 

```
Command line: xia2 image=/Users/graeme/data/i04-eiger-small/Therm_6_2_000001.h5
Provided input files not master files:
  /Users/graeme/data/i04-eiger-small/Therm_6_2_000001.h5
do you mean one of:
  /Users/graeme/data/i04-eiger-small/Therm_6_1_master.h5
  /Users/graeme/data/i04-eiger-small/Therm_6_2_master.h5
or:
  /Users/graeme/data/i04-eiger-small/Therm_6_2.nxs
  /Users/graeme/data/i04-eiger-small/Therm_6_1.nxs
```

Done in last commit 🙂

Applications/xia2_main.py
graeme-winter: Done

Applications/xia2_main.py
```diff
@@ -63,6 +65,36 @@ def check_environment():
...
+    for filename in master_files:
+        with h5py.File(filename, "r") as f:
+            if b"/data" in f and b"/entry" not in f:
+                bad.append(filename)
+
+    if bad:
+
+        dirs = set([os.path.split(b)[0] for b in bad])
+        masters = sum((glob.glob(os.path.join(d, "*_master.h5")) for d in dirs), [])
+        nxss = sum((glob.glob(os.path.join(d, "*.nxs")) for d in dirs), [])
```
rjgildea, line 84: I wonder if using `pathlib` here would make the code simpler?

Also I would probably prefer `itertools.chain.from_iterable()` to `sum`. Using `sum()` in this way seems like a bit of python dark magic.

rjgildea: Furthermore, `sum(list_of_lists, [])` has quadratic complexity, whereas `itertools.chain.from_iterable(list_of_lists)` has linear complexity:
https://mathieularose.com/how-not-to-flatten-a-list-of-lists-in-python/

graeme-winter: 🤷‍♂️ 

Humbly assert that the complexity of `sum` on a short list of short lists is unlikely to have an impact on how long it will take to compose the error message

But I suppose that this is _technically correct_

graeme-winter: Done as instructed. 

Personally I find this much more cumbersome and much less easier to read

Applications/xia2_main.py
```diff
@@ -63,6 +65,36 @@ def check_environment():
...
+
+        message = (
+            "Provided input files not master files:\n  "
+            + "\n  ".join(bad)
+            + "\ndo you mean one of:\n  "
+            + "\n  ".join(masters)
+            + "\nor:\n  "
+            + "\n  ".join(nxss)
```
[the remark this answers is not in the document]
graeme-winter: I think this could get people adding both nxs and master which would make for different problems, but I have made the change anyway.

Applications/xia2_main.py
```diff
@@ -63,6 +66,38 @@ def check_environment():
...
+    bad = []
+
+    for filename in master_files:
+        with h5py.File(filename, "r") as f:
+            if b"/data" in f and b"/entry" not in f:
+                bad.append(filename)
+
+    if bad:
+
+        dirs = set([os.path.split(b)[0] for b in bad])
```
Anthchirp, line 83:
```suggestion
        dirs = set(os.path.split(b)[0] for b in bad)
```

Applications/xia2_main.py
```diff
@@ -63,6 +66,38 @@ def check_environment():
...
+def check_hdf5_master_files(master_files):
+    """Check the input HDF5 master files look a little bit like HDF5 master
+    files and not just the data files: if the latter then sys.exit() with a
+    helpful message"""
+
+    bad = []
+
+    for filename in master_files:
+        with h5py.File(filename, "r") as f:
+            if b"/data" in f and b"/entry" not in f:
```
Anthchirp, line 78: Not sure about the wider context of this function. Do we already know that h5py can read these files or should we also catch any exceptions here?
(If in doubt: leave as is and sort out later)

Anthchirp:
I agree that `pathlib` may be useful here. However I don't have the magic spells at hand. We can always rewrite it after we decide to use `pathlib` over string based path processing.

Applications/xia2_main.py
graeme-winter: Guess it would be "bad" if `h5py` can't read so ... yeh will catch

graeme-winter: Catching `OSError` seems to get most everything

graeme-winter:
👆 waiting fo' the ol' green tick before squash merge

PR #514, opened 2020-08-24
Run test in a temp dir to avoid PHIL droppings

benjaminhwilliams:
This test creates the files xia2-diff.phil and xia2-working.phil.

### DIFF
--- a/Test/Schema/test_Schema.py
+++ b/Test/Schema/test_Schema.py
@@ -53,7 +53,7 @@ def test_load_imageset_template_missing_images(insulin_with_missing_image):
         assert imagesets[0].get_array_range() == (0, 22)
 
 
-def test_load_imageset_split_sweep(dials_data):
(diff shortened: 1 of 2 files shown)


rjgildea (approved):
Not sure newsfragment strictly necessary here (i.e. this is almost certainly not release note worthy), but probably doesn't cause any harm.

benjaminhwilliams:
I have  feeling `misc` news doesn't get transliterated into the `CHANGELOG.rst`, it just lists the PR/issue number.

benjaminhwilliams:
I might be wrong.  I'm assuming the configuration of Towncrier for xia2 is like that for screen19.

Anthchirp:
the towncrier for xia2 will be following the schema for DIALS, ie. https://github.com/dials/dials/blob/master/newsfragments/README.MD

PR #515, opened 2020-08-24
Replace load_imageset tests with xia2.setup tests

rjgildea:
Directly testing the load_imageset function is fragile as there are a
number of singletons lurking in the depths of xia2 which can mean shared
state between tests. More reliable to test this behaviour in a separate
process via the xia2.setup command.


benjaminhwilliams (commented):
This looks good but when I try to test it locally, I seem not to have a `xia2.setup` executable.  It exists in a recent CCP4 distribution of DIALS but not in my development build.  Am I missing something?

rjgildea:
> Am I missing something?

`libtbx.refresh`

benjaminhwilliams:
Thanks, that works.

This looks good to me, let's see what Jenkins makes of it.

benjaminhwilliams (approved):

PR #518, opened 2020-08-26
set_batch_offset() on imageset version of scan

rjgildea:
Apparently `expt.imageset.get_scan()` is a different scan instance to `expt.scan`. As highlighted by dials/dials#1383 it is necessary to ensure that both scan objects have the same batch_offset.


rjgildea:
Tests will currently fail as there is a dependency on https://github.com/dials/dials/pull/1383

PR #521, opened 2020-09-01
Update black to 20.8b1 and clean clutter

ndevenish:
Since the automatic job has failed for the last few weeks, here's a manual run with the new black changes.

I manually tweaked one bad-split where it treated a single argument function with a trailing comma as something that should be split.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
 
 # Automatic source code formatting
 - repo: https://github.com/psf/black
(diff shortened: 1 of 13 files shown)


Anthchirp (approved):
Looks fine, but would like to note that

> Since the automatic job has failed for the last few weeks

the automatic job has actually only failed once :wink:

ndevenish:
Ah, I saw several jobs failed and assumed it was just the once-per-week run.

PR #528, opened 2020-10-21
DC1: turn xia2 into a python package

Anthchirp:
This
* creates a static `setup.py` for xia2
* reorganises the repository into a source layout, ie. the xia2 package now lives in `/src/xia2`
* sets up a `libtbx_refresh.py` that `pip install -e`s xia2
* deals with version numbers at `libtbx_refresh`/`pip install` time, in a way that we should be able to use `bump2version` to make releases
* ~~creates a dispatcher for `xia2` only (rest are missing at this time)~~


Anthchirp:
One question is how we want to deal with the `command_line/` directory and how magic that should be.

We could keep up to 90% of the current libtbx-level of magic, ie. everything present becomes a dispatcher, names are decided by the libtbx magic comments, and the only change would be to settle on a common `run()` function.

Or we could go to the polar opposite and turn `command_line` into a regular directory and depend on developers to manually register commands in `setup.py`

Or possibly somewhere in-between.

graeme-winter:
> One question is how we want to deal with the `command_line/` directory and how magic that should be.
> 
> We could keep up to 90% of the current libtbx-level of magic, ie. everything present becomes a dispatcher, names are decided by the libtbx magic comments, and the only change would be to settle on a common `run()` function.
> 
> Or we could go to the polar opposite and turn `command_line` into a regular directory and depend on developers to manually register commands in `setup.py`
> 
> Or possibly somewhere in-between.

I'm not hostile to moving completely away from `magictbx` stuff... provided we kept some notes on how to add new things in $FUTURE

Anthchirp:
My current thinking is that we should abandon all *tbx magic and rename the `command_line/` directory to `cli/` to indicate a clear break between the philosophies.
The future way to make a command line tool is then to add a script name pointing to a function to `setup.py` like so:

Anthchirp:
Everything seems to work nicely.

The only directly noticeable effect is that for new dispatcher commands a total of 2 rounds of either 'libtbx.refresh' or 'libtbx.configure' or 'make reconf' need to be called. Possibly followed by `hash -r` if you tried to call your new command before completing the 2 rounds. Nothing can be really done about that as long as we are inside a cctbx environment.

`bumpversion` can be used to progress from
3.2.dev `--release-->` 3.2.0 `--patch-->` 3.2.1 `--patch-->` 3.2.2
and
3.2.dev `--minor-->` 3.3.dev `--release-->` 3.3.0 `--minor-->` 3.4.dev

So basically:
* `master`/`main` stays on `3.x.dev` until release time
* at which point we call `bump2version release`, making it `3.x.0`. This can be done on `master`/`main` or the release branch. It's probably more sensible to do this on the release branch.
* The release branch then progresses through `bump2version patch` for point releases.
* `master`/`main` on the other hand gets a `bump2version minor` to end up on `3.(x+1).dev` and normal development continues.

We may want to review the reported version string when on other branches - currently `xia2 --version` will report the commit depth and branch name, with this PR this will no longer happen - but that is a minor change, not urgent, and can be done separately.

Effects of file moving means `git log` will not be useful, but `git log --follow` will be. `git blame` will be fine. `git merge` with existing branches and files in old locations will work. This is for git 2.16.5 (standard RHEL7).

I would advise to `git clean -dffx` in the repository to remove any dangling `.pyc` files, but that should not be required.

This is ready for review.

Anthchirp:
Discussed in https://dials.github.io/kb/core/20201125#xia2-dc1

ndevenish:
This basically looks good to me.

I encountered a case where the `_show_xia2_version()` part of `libtbx_refresh.py` didn't succeed, but think this is just some minor race condition that I couldn't reproduce.

I also observed that with `pip install -e .`, shims were created in `conda_base/bin/xia2*`, whereas this didn't appear to happen when run through `libtbx_refresh.py`, but I assume there's something in place suppressing their generation in the special refresh environment.

Neither of these are reasons to avoid moving forwards.

ndevenish (approved):

Anthchirp:
Not quite sure what you saw there :thinking: 

Starting with a clean bootstrap from `master`, then moving the xia2 branch to `dc1` and running one `make reconf` the xia2 dispatchers in `build/bin` disappear, the ones in `conda_base/bin` are generated:
```
Processing: "/scratch2/wra62962/files/temp/bootstrap/modules/xia2/libtbx_refresh.py"
Obtaining file:///scratch2/wra62962/files/temp/bootstrap/modules/xia2
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Installing collected packages: xia2
  Running setup.py develop for xia2
Successfully installed xia2
Can't tell xia2 version
Processing: "/scratch2/wra62962/files/temp/bootstrap/build/exe_dev"
Done.
```
This gives you the error output of the show version call.

A subsequent `make reconf` restores the `build/bin` dispatchers with output
```
Processing: "/scratch2/wra62962/files/temp/bootstrap/modules/xia2/libtbx_refresh.py"
Obtaining file:///scratch2/wra62962/files/temp/bootstrap/modules/xia2
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Installing collected packages: xia2
  Attempting uninstall: xia2
    Found existing installation: xia2 0.7.121-g2f7d9a94-dc1
    Uninstalling xia2-0.7.121-g2f7d9a94-dc1:
      Successfully uninstalled xia2-0.7.121-g2f7d9a94-dc1
  Running setup.py develop for xia2
Successfully installed xia2
XIA2 0.7.121-g2f7d9a94-dc1
Processing: "/scratch2/wra62962/files/temp/bootstrap/build/exe_dev"
Done.
```

Another clean bootstrap with the xia2 directory on `dc1` from the start also gives the correct dispatchers in both locations.

ndevenish:
I've realised what happened with the dispatchers, they are put into `conda_base/python.app/Contents/bin/` on conda on mac, it seems, so not an issue here.

I can reproduce the `Can't tell xia2 version` by doing `pip uninstall xia2; libtbx.refresh`.

Anthchirp:
@rjgildea reported
```
  File "<string>", line 39, in <module>
  File "<string>", line 30, in _show_xia2_version
ImportError: bad magic number in 'xia2': b'\x03\xf3\r\n'
make: *** [reconf] Error 1
```

which was caused by `.pyc` files left over in the xia2 module directory, and could be fixed by `git clean -dffx` (note: this will remove all files and directories that are not part of the repository, so use with caution)

PR #529, opened 2020-11-03
Expose scaling refl selection method

rjgildea:
For narrow wedges, the default `dials.scale` option of `reflection_selection.method=quasi_random` can occassionally fail. Setting it to `reflection_selection.method=use_all` can work around this error. This PR exposes the parameter through xia2.multiplex.
Move the `Isigma_range` parameter inside the new `reflection_selection` phil scope.

### DIFF
--- a/Modules/MultiCrystal/ScaleAndMerge.py
+++ b/Modules/MultiCrystal/ScaleAndMerge.py
@@ -69,14 +69,29 @@ scaling
     .type = choice
   outlier_rejection = simple *standard
     .type = choice
-  Isigma_range = 2.0,100000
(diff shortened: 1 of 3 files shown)


jbeilstenedmands:
It's fine to expose these methods here as an extra option, however the default option `reflection_selection.method=quasi_random` seems to only fail if extra filters are added by the user (human or program), for example setting an `Isigma_range=` (or for very narrow wedges the default `min_partiality=0.95` option in dials.scale might also cause this issue). This being said, the `use_all` method is probably the most appropriate for very narrow wedges.

Issue #534, opened 2020-11-05
Run dials.export format=mmcif compress=bz2 automatically


dagewa:
As [promised](https://github.com/dials/dials/issues/1457#issuecomment-721816985) I'm creating an issue here about writing out the deposition file-of-record (mmcif) from xia2 runs. With https://github.com/dials/dials/pull/1482 merged this could be compressed to save disk space.

jbeilstenedmands:
I'm going to start working on this now, unless anyone else has already started doing so in secret 🙂

dagewa:
I was waiting for dials/dials#1482 to go in first, but there isn't really any reason to hold off. You know better where to look in xia2 than I do, so you're welcome to it ;-)

One thing I did want to try was to compare the filesizes of unmerged MTZ, uncompressed mmcif and the two supported compression modes for mmcif for some reasonably large unmerged case. Perhaps also the speed of writing the compressed files. Although the title suggests to use `gz`, the `bz2` file might be significantly smaller.

dagewa:
An experiment with a fairly meaty dataset (from `/dls/metrix/metrix/data_analysis/for_METRIX/results_20190326/xia2_stresstest/4L0N/`) shows a few interesting things:
```
fcx32934@zdar:/tmp$ time dials.export scaled.expt scaled.refl > /dev/null

real	0m16.622s
user	0m10.640s
sys	0m5.725s
fcx32934@zdar:/tmp$ time dials.export scaled.expt scaled.refl format=mmcif > /dev/null

real	2m37.118s
user	2m29.220s
sys	0m7.862s
fcx32934@zdar:/tmp$ time dials.export scaled.expt scaled.refl format=mmcif compress=gz > /dev/null

real	4m14.879s
user	4m7.603s
sys	0m7.121s
fcx32934@zdar:/tmp$ time dials.export scaled.expt scaled.refl format=mmcif compress=bz2 > /dev/null

real	2m48.300s
user	2m42.135s
sys	0m6.161s
fcx32934@zdar:/tmp$ ll -h scaled.mtz scaled.cif scaled.cif.gz scaled.cif.bz2
-rw-r--r-- 1 fcx32934 fcx32934 280M Nov  6 17:15 scaled.cif
-rw-r--r-- 1 fcx32934 fcx32934  87M Nov  6 17:23 scaled.cif.bz2
-rw-r--r-- 1 fcx32934 fcx32934 103M Nov  6 17:20 scaled.cif.gz
[...]

PR #537, opened 2020-11-09
Extend xia2.mmcif with unmerged reflection mmCIF data from DIALS

jbeilstenedmands:
For #534

Add the mmCIF block created by `dials.export format=mmcif` to the `xia2.mmcif` file.

TODO: Also add unmerged data to xia2.mmcif for `scaler = ccp4a, 3dii` ?


jbeilstenedmands:
For adding mmcif data for non-dials pipelines, my thought was to extract as much information as possible from the scaled_unmerged.mtz output files. GEMMI is easily able to do this (https://project-gemmi.github.io/wasm/to_mmcif/), so is it time to revisit https://github.com/dials/dials/issues/1266 ?
We would retain the ability to write more information from DIALS processing direct to mmcif, however the final output mmcif files may look different from dials vs non-dials pipelines.

dagewa:
I'm very much in favour of using Gemmi for all things cif. It is very fast and well documented.

jbeilstenedmands:
Okay, so I think this now works* for all pipelines, subject to some minor polishing, and should correctly handle MAD and multi-sweep datasets.

For pipeline=dials, this just picks up the mmcif from dials.export. For 3dii or dials-aimless, data is picked up from the mtz files and written into sections based on PROJECT_CRYSTAL_WAVE - so each of these sections has the merging stats and the unmerged data, and a PROJECT_CRYSTAL section contains the metadata regarding the physical crystal, scan and wave relationships.

I may look at the way dials.export writes mmcif data to see if I can make this more consistent with the layout here.

*3dii pipeline currently doesn't write the correct scan angle range as this is missing from mtz batch headers.

graeme-winter:
@jbeilstenedmands please can we make the CIF output _optional_? I don't mind the default being that we do output it, but if ${USER} wants to process some data in a hurry and has no interest in mmCIF output it seems wasteful to make such output mandatory

dagewa:
Note I recommend `bz2` compression for these output files to avoid wasting bytes. Also note that mmCIF output is still rather slow (see https://github.com/dials/dials/issues/1485 for some tests).

jbeilstenedmands:
An update; this is now in better shape, with a consistently formatted mmcif output across the pipelines.

The output contains a data_xia2 block which contains software information and citations and some overall information like two theta refined cell, symmetry and wavelength information. Then there is a data block for each 'WAVE' i.e. data_AUTOMATIC_DEFAULT_NATIVE or data_AUTOMATIC_DEFAULT_WAVE1, data_AUTOMATIC_DEFAULT_WAVE2 etc. Each of these has cell and scan information as well as the merging statistics and unmerged data for that wavelength.

This PR also does some tidying up whilst there, notably removing unused citations. I have also made a versions custom dictionary, to determine program versions only when requested. I don't expect people to look over all this code, but would be useful to know that people are happy with the output mmcif file.

jbeilstenedmands:
Think has a few more changes on the way, which will be done by the end of today/this weekend ready for dials 3.3

PR #539, opened 2020-11-13
Use anomalous scaling for pipeline=dials

jbeilstenedmands:
So far, in the dials pipeline, we have been following the Aimless approach of not separating anomalous pairs for scaling, however for strong anomalous data this seems to be having a negative effect on the output anomalous statistics. So this PR changes the behaviour and sets anomalous=True in dials.scale if setting atom=X/anomalous=True. This brings the behaviour of the dials pipeline more in line with the 3dii pipeline.

As a test, I have taken 9 strongly diffracting datasets from I23, initially processed with xia2-dials and rescaled with anomalous=True/False, also shown is the 3dii processing as comparison. cc_anom shows that we are clearly suppressing some of the anomalous signal if anomalous=False for many of these datasets. 
I23_cc_anom_overall
I23_cc_anom_lowres


graeme-winter:
@jbeilstenedmands new pls 🙂

graeme-winter:
Will run some tests

Modules/Scaler/DialsScaler.py
```diff
@@ -89,6 +89,9 @@ def _updated_dials_scaler(self):
             PhilIndex.params.dials.scale.partiality_threshold
         )
 
+        if self.get_scaler_anomalous():
```
graeme-winter, line 92: Could oneline this by `self._scaler.set_anomalous(self.get_scaler_anomalous())`?

graeme-winter (approved):
Change set looks sane

Approve modulo:
 - evaluation
 - news

ndevenish:
I'm intending to merge this on Monday for DIALS 3.3.

ndevenish:
Apparently we marked this for further testing before merge - I'd swear that's the reason and that I didn't forget about it, except I have forgotten whether that's the case or not.

After discussion offline we've agreed that it should get more testing, whether this is as a post-release merge or not is TBD.

jbeilstenedmands:
I have been testing this change on a number of test datasets (selected from the work for the metrix analysis), which I have detailed in a separate document on the DLS sharepoint site. https://dlsltd.sharepoint.com/:f:/s/DIALS/EjJFcEzZCJhIv5g0wEJYA3QBv-iu1kWioU1qiz22nhleIw?e=jy6nrm
I have been comparing processing with 3dii, dials (main) and dials (this PR), and done experimental phasing with crank2 and/or phenix.autobuild.

A summary of the findings so far:
In terms of the merging statistics, switching between anomalous=True/False for scaling in dials does not result in any meaningful changes for some datasets where cc_anom is already quite high, whereas for other datasets the cc_anom is significantly increased from low/no cc_anom if anomalous=False.
In terms of phasing success, I have 3 datasets (Thaumatin, Ferritin, Insulin) where I was able to build a structure with all 3 xia2 processing options, and 2 where none of the xia2 processing options worked (Insulin, Lysozyme).

My conclusion is therefore that these changes should go ahead, as the strength of the anomalous signal is generally improved. While I have not found an example where this makes the difference between phasing success and failure, there was no detrimental effect to using anomalous=True. I believe this is consistent with what was previously observed for the metrix project, where the anomalous signal from xia2-dials statistics seems lower than it is in reality, given how many can successfully be phased.

graeme-winter:
Is this too substantial for a backport?

jbeilstenedmands:
The code changes themselves are not substantial, but given this is a major change in processing I was thinking we'd release in v3.5?

rjgildea (approved):
Sounds like this is a worthwhile change 👍

newsfragments/539.feature
```diff
@@ -0,0 +1 @@
+For the dials scaler, use anomalous scaling if atom=X/anomalous=True specified.
```
rjgildea, line 1:
```suggestion
Separate anomalous pairs when scaling with `dials.scale` if `anomalous=True`.
```

PR #547, opened 2020-12-10
Add a Github Action to check for newsfragments

ndevenish:
This Github Action runs on pull requests, and does two things:

- Checks for a newsfragment (anything added in `newsfragments/`). It will fail if this is not present.
- Looks for `newsfragments/XXX.*` and renames them to `newsfragments/<PR number>.*`. This means that you can create a PR with a placeholder newsfragment without having to go back to the repository, edit and push a new one.

~This PR will fail the check, because it does not yet have a newsfragment.~ TBD


Anthchirp:
Notification mail:
> DiamondLightSource-build-server pushed 1 commit.
>  48c8a64 Rename newsfragments/XXX.misc to newsfragments/547.misc

Could we change the commit author to indicate that this comes in via the build server account? I'm thinking this could avoid some confusion in the long term.

graeme-winter:
Approve of adding this, think will be a useful addition

I particularly like the `XXX.bugfix` (or whatever) news feature

ndevenish:
> Could we change the commit author to indicate that this comes in via the build server account? I'm thinking this could avoid some confusion in the long term.

Ah, I had forgotten about notification emails. In all other respects, it seemed like it was entirely hidden which user was doing the changes. Happy to change it.

Anthchirp (approved):

PR #549, opened 2020-12-16
DIALS 2.2.11

Anthchirp:
* fix and deprecate xia2.chef


benjaminhwilliams:
Perfect, thanks.  That's more-or-less byte-for-byte the solution I was drafting.

PR #550, opened 2020-12-16
Allow newsfragments Action to rename any sequence [xX]+

ndevenish:
Given there were comments complaining about FORTRAN case this seems like it'd be easier to remember.

Also, handle case where rename would cause conflict.


Anthchirp:
I would prefer to take any files added in the pull request that do not match `^\d+\.` and rename those rather than making allowances for different numbers and cases of `x`s.

Anthchirp:
The required logic for this could likely be similar to https://github.com/xia2/xia2/blob/dials-2.1/.travis/run-flake8 (`$TRAVIS_BRANCH` denotes the target branch name)

ndevenish:
I mean, the different numbers and cases of `X` are mainly to provide some robustness over remembering exactly what it is, for convenience, and flexibility - I suppose it's okay for xia2 as no push-to-master but there have been several times I've wanted to make newsfragments out of a commit sha directly, when someone has forgotten to put one in and I wanted to mark the commit for releases. It also makes what you are supposed to do clear "Write an XXX.bugfix" vs "Write a .bugfix extension file with the name of your choice" seems needlessly complicated.

I just 100% want to avoid a situation where someone is having to inadvertantly try and undo the change this has made.

I'm not so wedded to the restriction that I would complain if you wanted to change it.

ndevenish:
We can always change this later!

PR #551, opened 2020-12-16
Add dano/sigdano by resolution plot to xia2.html

jbeilstenedmands:
This is the xia2 counterpart to https://github.com/dials/dials/pull/1483

If anomalous=True, this adds the dano/sigdano plot to each 'Dataset X' tab in the analysis by resolution section.

This will fail the tests until the DIALS PR is merged.

The `dials.merge.html` report is suppressed as we are not using dials.merge to do the truncation, so no plot will be generated there.

### DIFF
--- a/src/xia2/cli/xia2_html.py
+++ b/src/xia2/cli/xia2_html.py
@@ -9,10 +9,15 @@ import traceback
 from collections import OrderedDict
 
 from libtbx import phil
+from iotbx import mtz
 import xia2
 from xia2.Modules.Report import Report
 from xia2.Handlers.Citations import Citations
 import xia2.Handlers.Streams
+
+from math import isclose
+from dials.algorithms.merging.merge import make_dano_plots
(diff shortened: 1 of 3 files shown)


src/xia2/cli/xia2_html.py
```diff
@@ -201,6 +206,25 @@ def generate_xia2_html(xinfo, filename="xia2.html", params=None, args=[]):
...
             }
 
+            anom_data = {}
+            mtz_file = mtz.object(file_name=reflection_files["mtz"])
+            # Collect F+, F-, SigF+, SigF- data for the correct wavelength
+            for array in mtz_file.as_miller_arrays():
+                if (
+                    len(array.info().labels) == 4
+                    and array.info().type_hints_from_file == "amplitude"
+                    and isclose(
```
rjgildea, line 216: TIL: `math.isclose`

rjgildea (approved):

PR #554, opened 2021-01-04
Bump version: 3.3.dev → 3.4.dev

ndevenish:
This is the version bump to 3.4.dev now that the [dials-3.3](https://github.com/xia2/xia2/tree/dials-3.3) branch has been split off and we're supposed to be using [the new](https://github.com/xia2/xia2/pull/528#issuecomment-733792471) release scheme. Since it's the first release using the new versioning scheme, I thought wider examination might be worthwhile.

Along with this merge, I will add an annotated tag to this commit for `v3.4.dev`. This means that xia2 version will self-describe roughly in the form:
```
$ xia2 --version
XIA2 3.3.dev-14-gf8877135-remotes-origin-HEAD
DIALS 3.dev.278-gc3447f3de
```
however - we probably want to change the version generation to do `3.3.dev14[-...]` because that appears to be the more correct [PEP 440](https://www.python.org/dev/peps/pep-0440/#developmental-releases) form. My xia2 still self-reports as `XIA2 0.7.137-gf8877135-remotes-origin-HEAD` - I think the intention was to migrate the version scheme with https://github.com/xia2/xia2/pull/528 but it looks like the tag part was left out.

And yes, I'm aware this adds a trailing space. As far as I can tell, this appears to be a limitation in python's [`configparser`](https://docs.python.org/3/library/configparser.html) writing and not part of bump2version.


Anthchirp (approved):
That's fine. The details on the version string stuff are still on my todo list somewhere.

ndevenish:
And I've tagged the branch now

Issue #292, opened 2019-04-12
xia2 3dii can't cope with more than 10M integrated reflections


rjgildea:
Parsing of the INTEGRATE.LP fails, because it assumes white space separation in output:
https://github.com/xia2/xia2/blob/master/Wrappers/XDS/XDSIntegrateHelpers.py#L75
I assume these are fixed-width columns, so we should treat as such.

 ```
******************************************************************************
                     PROCESSING OF IMAGES     2601 ...    2700
 ******************************************************************************

 USING  16 PROCESSORS

 *** DEFINITION OF SYMBOLS ***
 IER     = ERROR CODE AFTER ACCESSING DATA IMAGE
              0: NO ERROR
             -1: CANNOT OPEN OR READ IMAGE FILE
             -3: WRONG DATA FORMAT
 SCALE   = SCALING FACTOR FOR THIS DATA IMAGE
 NBKG    = NUMBER OF BACKGROUND PIXELS ON DATA IMAGE
 NOVL    = NUMBER OF OVERLOADED REFLECTIONS ON DATA IMAGE
 NEWALD  = NUMBER OF REFLECTIONS CLOSE TO THE EWALD SPHERE
 NSTRONG = NUMBER OF STRONG REFLECTIONS ON DATA IMAGE
 NREJ    = NUMBER OF UNEXPECTED REFLECTIONS
 SIGMAB  = BEAM_DIVERGENCE_E.S.D.=SIGMAB
 Parsing of the INTEGRATE.LP fails:

SIGMAR  = REFLECTING_RANGE_E.S.D.=SIGMAR (MOSAICITY)

 IMAGE IER  SCALE     NBKG NOVL NEWALD NSTRONG  NREJ   SIGMAB   SIGMAR
  2601   0  0.975 16680318    01092650     283     8  0.01281  0.40765
  2602   0  0.969 16680543    01092588     268     6  0.01308  0.46286
  2603   0  0.973 16680424    01092451     274     5  0.01419  0.44668
  2604   0  0.979 16679844    01092457     282     7  0.01460  0.43985

```

PR #564, opened 2021-02-02
Fix missing SENSOR_THICKNESS= for EIGER

rjgildea:
Bug introduced by #292

See also cctbx/dxtbx#296

* add end-to-end xia2 vmxi_thaumatin eiger test

* Different fallback PILATUS/EIGER sensor thicknesses


src/xia2/Wrappers/XDS/XDS.py
```diff
@@ -253,18 +253,17 @@ def imageset_to_xds(
...
-            if not thickness:
-                thickness = 0.32
-                logger.debug(
-                    "Could not determine sensor thickness. Assuming default PILATUS 0.32mm"
-                )
         except Exception:
-            thickness = 0.32
+            thickness = None
+        if not thickness:
+            thickness = {"PILATUS": 0.32, "EIGER": 0.45}.get(detector)
```
graeme-winter, line 262: Without this line here, does the test above use the correct sensor thickness? i.e. we _do_ get the right value from the header and setting the default here is just for convenience.

src/xia2/Wrappers/XDS/XDS.py
```diff
@@ -253,18 +253,17 @@ def imageset_to_xds(
...
             logger.debug(
-                "Error occured during sensor thickness determination. Assuming default PILATUS 0.32mm"
+                "Could not determine sensor thickness. Assuming default %s %.2f mm",
```
graeme-winter, line 264: I would have f-string'd here

src/xia2/Wrappers/XDS/XDS.py
graeme-winter: The correct value is present:

```
Grey-Area cctbx :) $ h5ls -rvd /Users/graeme/svn/cctbx/build/dials_data/vmxi_thaumatin/image_15799_master.h5/entry/instrument/detector/sensor_thickness
Opened "/Users/graeme/svn/cctbx/build/dials_data/vmxi_thaumatin/image_15799_master.h5" with sec2 driver.
entry/instrument/detector/sensor_thickness Dataset {SCALAR}
    Attribute: units scalar
        Type:      3-byte null-terminated ASCII string
        Data:  "mm"
    Location:  1:15832
    Links:     1
    Storage:   8 logical bytes, 8 allocated bytes, 100.00% utilization
    Type:      native double
    Data:
        (0) 0.45
```

graeme-winter: Additional: CdTe has sensor thickness of 750µm not 450µm

https://www.dectris.com/products/eiger2/eiger2-x-xe-cdte-for-synchrotron/eiger2-x-cdte-4m/

graeme-winter (approved):
Good fix, some questions raised re: defaults which may need some though.

src/xia2/Wrappers/XDS/XDS.py
rjgildea: The try/except fallback to defaults was there already - I just updated it to differentiate between PILATUS and EIGER. I'm not sure under what circumstances we would get an exception here, but presumably it is there for a reason...

graeme-winter: Not unheard of for this to be absent from headers because reasons -> error -> default is better estimate than 0...

I did wonder if we should actually _remove_ this block and instead fail...

PR #569, opened 2021-02-25
Update magic "reflections" number in tests 🎩 🐇

ndevenish:
Due to changes from https://github.com/dials/dials/pull/1297 the normal, expected number of reflections has increased slightly. Increase that magic number so the test don't fail.

This currently fails with
```
>       assert (
            abs(mtz_object.n_reflections() - expected_reflections) < 300
        ), mtz_object.n_reflections()
E       AssertionError: 48482
E       assert 859 < 300
E        +  where 859 = abs((48482 - 47623))
E        +    where 48482 = <bound method n_reflections of <iotbx_mtz_ext.object object at 0x7f7982061430>>()
E        +      where <bound method n_reflections of <iotbx_mtz_ext.object object at 0x7f7982061430>> = <iotbx_mtz_ext.object object at 0x7f7982061430>.n_reflections
```

### DIFF
--- a/newsfragments/569.bugfix
+++ b/newsfragments/569.bugfix
@@ -1,0 +1,1 @@
+Fix tests affected by changes to profile fitting in `dials/dials#1297 <https://github.com/dials/dials/pull/1297>`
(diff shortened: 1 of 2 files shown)


rjgildea (approved):
We know there is an upstream reason for the change in number of reflections, so this change makes sense in the context of the test 👍

PR #570, opened 2021-03-04
Workaround missing ignore flags

### LINKED ISSUE #567: xia2 crashing with 'FormatHDF5 has no attribute ignore'
```
Traceback (most recent call last):
  File "dials-v3-3-2/modules/xia2/src/xia2/cli/xia2_main.py", line 328, in run
    xia2_main()
  File "dials-v3-3-2/modules/xia2/src/xia2/cli/xia2_main.py", line 50, in xia2_main
    CommandLine = get_command_line()
  File "dials-v3-3-2/modules/xia2/src/xia2/Applications/xia2_main.py", line 145, in get_command_line
    write_xinfo(xinfo, directories)
  File "dials-v3-3-2/modules/xia2/src/xia2/Applications/xia2setup.py", line 623, in write_xinfo
    sweeps = _rummage(directories)
  File "dials-v3-3-2/modules/xia2/src/xia2/Applications/xia2setup.py", line 585, in _rummage
    templates.update(visit(root, files))
  File "dials-v3-3-2/modules/xia2/src/xia2/Applications/xia2setup.py", line 205, in visit
    elif format_class.ignore():
AttributeError: type object 'FormatHDF5' has no attribute 'ignore'
```

This is caused by cctbx/dxtbx#255 which didn't account for necessary changes in xia2.

graeme-winter:
Temporary patchup until this is universal

Fixes #567

Co-authored-by: Markus Gerstel <markus.gerstel@diamond.ac.uk>


src/xia2/Applications/xia2setup.py
```diff
@@ -202,7 +202,11 @@ def visit(directory, files):
                     "Ignoring %s (Registry can not find format class)" % full_path
                 )
                 continue
-            elif format_class.ignore():
+            elif getattr(
+                format_class, "ignore", getattr(format_class, "is_abstract")
+            )():
+                # temporary workaround while not all format classes have the
+                # ignore method - https://github.com/xia2/xia2/issues/567
```
ndevenish, line 209: Do we have any cases where xia2 will be encountering `is_ignore`, or classes without `is_abstract()`? Or just defensive check?
```suggestion
                # temporary workaround while not all format classes use the
                # is_abstract method - https://github.com/xia2/xia2/issues/567
```

graeme-winter: Not certain I agree - this gets the `ignore()` method first, and if that is `None` gets the `is_abstract()` method. 

Ah, because the code change is wrong -> fixing.

src/xia2/Applications/xia2setup.py
```diff
@@ -202,7 +202,11 @@ def visit(directory, files):
                     "Ignoring %s (Registry can not find format class)" % full_path
                 )
                 continue
-            elif format_class.ignore():
+            elif getattr(
+                format_class, "is_abstract", getattr(format_class, "ignore")
+            )():
+                # temporary workaround while not all format classes have the
+                # is_abstract method - https://github.com/xia2/xia2/issues/567
```
ndevenish, line 209:
```suggestion
            elif format_class.is_abstract():
```
So, given discussion, I think it's safe to not check ignore now? the only case I can think of that would not work here is if someone had a 
- custom format class
- that they used only with xia2. 

I think that set is probably zero?

PR #577, opened 2021-03-24
DIALS 3.4.1

DiamondLightSource-build-server:
Features
--------

- ``xia2.multiplex``: Use resolution cutoff determined during scaling for cluster analysis (#576)

Bugfixes
--------

- ``xia2.multiplex``: Fix cos-angle clustering varying between runs (#576)

### DIFF
--- a/src/xia2/Test/regression/test_multiplex.py
+++ b/src/xia2/Test/regression/test_multiplex.py
@@ -1,4 +1,6 @@
 import json
+import os
+import pathlib
 import pytest
 
 from dxtbx.model import ExperimentList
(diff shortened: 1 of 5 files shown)


src/xia2/Test/regression/test_multiplex.py
```diff
@@ -30,13 +32,21 @@
 ]
 
 
-def test_proteinase_k(mocker, regression_test, dials_data, tmpdir):
+@pytest.fixture()
+def proteinase_k(regression_test, dials_data, tmp_path):
     data_dir = dials_data("multi_crystal_proteinase_k")
     expts = sorted(f.strpath for f in data_dir.listdir("experiments*.json"))
     refls = sorted(f.strpath for f in data_dir.listdir("reflections*.pickle"))
+    os.chdir(tmp_path)
```
Anthchirp, line 40: this looks like asking for trouble

ndevenish: Hmm, I don't like it took much, it is a bit unexpected/implicity. OTOH, it does remove five `tmpdir.as_cwd()` and their respective indent blocks.

I wonder if putting the tmpdir context manager around the yield would be better?

PR #593, opened 2021-05-12
Fix documentation section on resolution estimation

rjgildea:


graeme-winter (approved):
Appreciate that thank you - wondered about including a specification for what the current defaults are, but realised that this would rapidly get out of date (again)

PR #597, opened 2021-06-06
Run tests on Azure in Python development mode

Anthchirp:
This currently causes tests to fail due to a dxtbx datablock `DeprecationWarning` showing up.

You can reproduce this easily with:
```bash
PYTHONDEVMODE=1 xia2.setup image=$(dials.data get insulin -q)/insulin_1_001.img read_all_image_headers=False >/dev/null
```

### DIFF
--- a/.azure-pipelines/unix-build.yml
+++ b/.azure-pipelines/unix-build.yml
@@ -77,7 +77,7 @@ steps:
     set -ux
     export DIALS_DATA=${PWD}/data
     cd modules/xia2
-    # export PYTHONDEVMODE=1
+    export PYTHONDEVMODE=1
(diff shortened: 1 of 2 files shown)


rjgildea:
This `DeprecationWarning` doesn't come from xia2 using `DataBlocks` directly, but because it calls `ExperimentList.from_templates()` which apparently still goes via the `DataBlockTemplateImporter()`:

https://github.com/xia2/xia2/blob/6d6d73003234b735d41866c253c49912472efc04/src/xia2/Schema/__init__.py#L156
https://github.com/cctbx/dxtbx/blob/main/model/experiment_list.py#L811-L820

Anthchirp:
Looks to be fixed with cctbx/dxtbx#381. Will wait for dials/dials#1752 to be merged before updating and re-running this PR

PR #598, opened 2021-06-08
Build on Python 3.9

Anthchirp:
we didn't do this historically, but thanks to #596 we can now afford it.


graeme-winter:
Python 3.9 build worked fine for me on macOS 11.4 (though I used system compilers rather than Conda ones because of a recent `xcode` update)

Moving to 3.9 seems like a sensible move, now 3.8 is in security patch only.

PR #603, opened 2021-07-07
xia2.multiplex: add absorption_level parameter for scaling.

jbeilstenedmands:
Exposes the `dials.scale` `absorption_level` parameter, with a default of `None`.
While we're here, also change the model setting so that we don't set the model explicitly if the model is `auto` - this
will let `dials.scale` turn on the absorption correction for large sweeps (>60 degrees due to `dials.scale` `auto` model rules).

Old behaviour:
`xia2.multiplex` on several large (>60 degree) sweeps:  no absorption correction surfaces used 🙁 
New behaviour:
`xia2.multiplex` on several large (>60 degree) sweeps: absorption correction surfaces used 🙂 

New behaviour once https://github.com/dials/dials/pull/1793 merged:
`xia2.multiplex` on any size sweeps + `absorption_level=x`: absorption correction surfaces used with absorption level set 😃


jbeilstenedmands:
It seems that the test failure is due to the fact that the scaling auto model rules are different if the scan oscillation range is < 5 degrees as is the case for the failing test. I think updating the dials code is most sensible here, so I'll do that and then this should pass.

PR #624, opened 2021-09-22
Saturated pixel checks

graeme-winter:
Add checks for overloaded pixels to the spot finding, as well as an alert that spots are overloaded in stdout:

```
-------------------- Spotfinding SWEEP1 --------------------
67135 spots found on 450 images (max 1261 / bin)
*                                         *                *
* ** ********* *********************************************
************************************************************
************************************************************
************************************************************
************************************************************
************************************************************
************************************************************
************************************************************
************************************************************
1                         image                          450
Overloads found: 11 spots / 11 pixels
------------------- Autoindexing SWEEP1 --------------------
All possible indexing solutions:
cI  77.86  77.86  77.86  90.00  90.00  90.00
hR 110.05 110.05  67.45  90.00  90.00 120.00
tI  77.81  77.81  77.65  90.00  90.00  90.00
oI  77.66  77.81  77.84  90.00  90.00  90.00
oF  77.65 110.01 110.05  90.00  90.00  90.00
mC 109.88  77.83  77.64  90.00 134.93  90.00
aP  67.38  67.32  67.31 109.36 109.46 109.59
Indexing solution:
cI  77.86  77.86  77.86  90.00  90.00  90.00
-------------------- Integrating SWEEP1 --------------------
```

Currently controlled by a PHIL parameter 

```
general.check_for_saturated_pixels=True
```

### DIFF
--- a/src/xia2/Modules/Indexer/DialsIndexer.py
+++ b/src/xia2/Modules/Indexer/DialsIndexer.py
@@ -6,6 +6,7 @@ import logging
 import math
 import os
 import string
+import time
 
 import libtbx
 from cctbx import crystal, sgtbx
@@ -321,6 +322,14 @@ class DialsIndexer(Indexer):
             gain = PhilIndex.params.xia2.settings.input.gain
             if gain:
                 spotfinder.set_gain(gain)
+
+            # set a limit for spot finding which is 25% greater than we
+            # actually trust - can use this to measure overloads
+
+            if PhilIndex.params.general.check_for_saturated_pixels:
+                count_limit = imageset.get_detector()[0].get_trusted_range()[1]
+                spotfinder.set_maximum_trusted_value(count_limit * 1.25)
+
             if dfs_params.filter_ice_rings:
                 spotfinder.set_filter_ice_rings(dfs_params.filter_ice_rings)
             if dfs_params.kernel_size:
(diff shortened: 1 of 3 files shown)


src/xia2/Modules/Indexer/DialsIndexer.py
```diff
@@ -321,6 +322,14 @@ def _index_prepare(self):
             gain = PhilIndex.params.xia2.settings.input.gain
             if gain:
                 spotfinder.set_gain(gain)
+
+            # set a limit for spot finding which is 25% greater than we
+            # actually trust - can use this to measure overloads
```
benjaminhwilliams, line 327: Is the choice of 25% an arbitrary factor in the right ballpark, or is there a specific motivation?

src/xia2/Modules/Indexer/DialsIndexer.py
```diff
@@ -346,6 +355,50 @@ def _index_prepare(self):
...
+                for j in range(nn):
+                    b = boxes[j]
+                    pixel_histogram.update(
+                        flex.histogram(
+                            b.data.as_double().as_1d(),
+                            data_min=0,
+                            data_max=1.25 * count_limit,
+                            n_slots=5,
+                        )
+                    )
```
benjaminhwilliams, line 382: Is this quicker than concatenating all the flattened shoebox pixel intensity values and simply calling `flex.histogram` (or `numpy.bincount` or whatever) on the single array?

src/xia2/Wrappers/Dials/Spotfinder.py
```diff
@@ -88,6 +89,9 @@ def set_filter_ice_rings(self, filter_ice_rings):
         def set_gain(self, gain):
             self._gain = gain
 
+        def set_maximum_trusted_value(self, maximum_trusted_value):
+            self._maximum_trusted_value = maximum_trusted_value
+
```
benjaminhwilliams, line 94: If this attribute is essentially a simple parameter, why do we bother with the setter at all?  Why not simply make the attribute public (i.e. `self.maximum_trusted_value`) and assign to it, rather than calling a method to do the same thing?

src/xia2/Modules/Indexer/DialsIndexer.py
graeme-winter: The former - though I noted in the few Eiger sets I looked at there were very few pixels > 5% higher than trusted limit. Would you like better justification for this?

I imposed a limit as e.g. relying on 0xffff or similar could pick up incorrectly assessed mask pixels.

src/xia2/Modules/Indexer/DialsIndexer.py
graeme-winter: My concern was more memory footprint rather than calculation time - though I have no data to assess this. 

I'll run with a big data set & record timing for comparison.

src/xia2/Wrappers/Dials/Spotfinder.py
graeme-winter: Your comment is reasonable and valid - however the "house style" is the getter / setter pattern & this is fitting in with that. 

The question of "is the house style wrong" would be a reasonable follow up.

src/xia2/Modules/Indexer/DialsIndexer.py
graeme-winter: Now running with a 9,000 image semi-saturated Eiger 16M data set to get worst-case timing added with current protocol.

graeme-winter: OK, for this case it took 

```
Overloads found: 5 spots / 5 pixels
Overload detection took 3.32s
```

which I think is probably OK given 

```
675716 spots found on 9000 images (max 11941 / bin)
```

Issue #626, opened 2021-09-22
Retire the "clever" Driver creation pattern


graeme-winter:
i.e. 

```
def AlignCrystal(DriverType=None):
    """A factory for AlignCrystalWrapper classes."""

    DriverInstance = DriverFactory.Driver(DriverType)

    class AlignCrystalWrapper(DriverInstance.__class__):
        def __init__(self):
            DriverInstance.__class__.__init__(self)
            self.set_executable("dials.align_crystal")

            self._experiments_filename = None
            self._json_filename = "align_crystal.json"
```

which is nothing but a mystery for `xia2` devs and not a thing which has (in 16 years) actually proved to be useful.

PR #627, opened 2021-09-24
Remove driver type concepts

graeme-winter:
To remove "ghosts" from the xia2 forest - start work towards an assertive fix for #626  since we don't have (as far as I am aware) downstream users of this API.


src/xia2/Driver/DriverFactory.py
```diff
@@ -12,21 +7,14 @@ def __init__(self):
...
-            self.set_driver_type(os.environ["XIA2CORE_DRIVERTYPE"])
-
     def set_driver_type(self, driver_type):
         """Set the kind of driver this factory should produce."""
         if driver_type not in self._implemented_types:
             raise RuntimeError("unimplemented driver class: %s" % driver_type)
+        else:
+            raise RuntimeError("setting driver_type now unsupported")
 
         self._driver_type = driver_type
```
rjgildea, line 19: It's probably better just to either add simply add a `DeprecationWarning` to the existing method, or replace the entire method with a single `RuntimeError`. The current approach results in a lgtm warning that the line `self._driver_type = driver_type` can never be hit.

rjgildea (commented):
Should we just remove `DriverFactory` entirely and make all wrappers inherit directly from `SimpleDriver`?

PR #631, opened 2021-11-04
xia2.multiplex: extend deltacchalf filtering options to match those in dials.scale

rjgildea:
Depends on https://github.com/dials/dials/pull/1913


jbeilstenedmands (approved):
Looks good, tested the new options and seems to work well, thanks for implementing 👍

PR #636, opened 2021-11-13
Use FuncFormatter to convert ds2 to d-spacings

### LINKED ISSUE #634: xia2.compare_merging_stats warning written to stderr
Currently getting this:
```
/home/fcx32934/sw/cctbx/modules/xia2/src/xia2/cli/compare_merging_stats.py:324: UserWarning: FixedFormatter should only be used together with FixedLocator
  ax.set_xticklabels(xticks_d)
```
I was puzzled because I was calling `xia2.compare_merging_stats` at the end of a BASH script that had `set -e`. It took a while to work out that the program wasn't running because the script was exiting... :facepalm: 

I'll have a look at this warning when I get a chance.

rjgildea:
Alternative to #635 

Fixes #634


dagewa (approved):
Works for me, and is a better solution than #635 :+1: . Some of the plots look a bit different as a result, but in no way worse as far as I can tell

PR #641, opened 2021-12-16
Multiplex imported mtz

dagewa:
Experiments imported by [`dials.import_mtz`](https://github.com/dials/dials/tree/import_mtz) might not have an imageset. In that case, don't fail when trying to access that object.

This does not yet allow `xia2.multiplex` to work on the output of `dials.import_mtz`, but it gets us further. The next issue is that `dials.two_theta_refine` fails, as the imported reflection table does not have `"xyzobs.mm.value"`.


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -248,8 +248,11 @@ def _set_batches(self):
 
         for i, expt in enumerate(self._experiments):
             expt.scan.set_batch_offset(i * 10 ** n)
-            # This may be a different scan instance ¯\_(ツ)_/¯
-            expt.imageset.get_scan().set_batch_offset(expt.scan.get_batch_offset())
+            try:
+                # This may be a different scan instance ¯\_(ツ)_/¯
+                expt.imageset.get_scan().set_batch_offset(expt.scan.get_batch_offset())
+            except AttributeError:
+                pass
```
rjgildea, line 255: Is this just catching the case where `expt.imageset is None`? If so, could this just be a simple `if` statement rather than `try`/`except`? 
```suggestion
            if expt.imageset:
                # This may be a different scan instance ¯\_(ツ)_/¯
                expt.imageset.get_scan().set_batch_offset(expt.scan.get_batch_offset())
```

rjgildea (approved):
Once multiplex is confirmed working with mtz-imported data, it would probably be a good idea to add a suitable test.

dagewa:
> Once multiplex is confirmed working with mtz-imported data, it would probably be a good idea to add a suitable test.

Agree. I have been discussing with @jbeilstenedmands about finding suitable test data for `dials.import_mtz`. Once we have that, perhaps it would make sense to use the same data sets imported to MTZ for a multiplex test. At the moment, `dials.import_mtz` might still change so we shouldn't fix on a particular MTZ result yet.

ndevenish:
Will this be a test here, or in DIALS

as in - is there any reason to hold this up for the DIALS pr, or safe to just merge now?

dagewa:
No reason to hold this up. There won't be a test added to this PR. There are likely to be tests added to DIALS and xia2 in future, but this is just a prerequisite change first

PR #647, opened 2022-01-19
More explicit error message

rjgildea:


graeme-winter (approved):
Thanks, thought the same

PR #651, opened 2022-01-26
Fix for xia2/xds hanging when run from ccp4i2

dagewa:
The original `procrunner` call to `pointless` causes xia2 to hang when started from ccp4i2. I don't know exactly why, but providing a timeout avoids that problem. Maybe there is a better way.

Removed `--no-input` as well. This does not appear to be a real `pointless` option.


src/xia2/Handlers/Versions.py
```diff
@@ -63,7 +63,12 @@ def get_aimless_version():
 
 def get_pointless_version():
     result = procrunner.run(
-        ["pointless", "--no-input"], print_stdout=False, print_stderr=False
+        [
+            "pointless",
+        ],
+        print_stdout=False,
+        print_stderr=False,
+        timeout=1,
```
graeme-winter, line 71: This is succeeding by simply _killing_ the pointless process as it waits for `stdin`

graeme-winter (approved):
Good find, makes sense (at least, answers the context)

The answer is not _nice_ but I can see that it works. I would imagine that this does not change the behaviour outside of the context of i2 -> LGTM

I do wonder if a comment as to _why_ this is there would be worth adding? "Timeout to prevent hanging of call in i2 environment" or similar.

Anthchirp:
Missing a raise_timeout_exception=False parameter, and leaves the question open why this happens in the first place.

Is there a way to reproduce this outside of i2? If no, how are the environments different?

src/xia2/Handlers/Versions.py
dagewa: what's weird is that `pointless` run from the command line does _not_ appear to wait, so I don't know why this happens in i2

dagewa:
> Missing a raise_timeout_exception=False parameter, and leaves the question open why this happens in the first place.
> 
> Is there a way to reproduce this outside of i2? If no, how are the environments different?

I don't know

dagewa:
This minimal example demonstrates the issue outside of i2 (I ran it with `ccp4-python` though, from a CCP4 8.0 test bundle, though that might not matter):
```
from PySide2 import QtCore

p = QtCore.QProcess()
p.start(
    "xia2",
    [
        "pipeline=3dii",
        'image="/home/fcx32934/data/betalac/C2sum_1_0001.cbf.gz:1:50"',
    ],
)
status = p.waitForFinished(-1)
print(status)
```

Anthchirp:
This smaller example also hangs indefinitely:
```python
from PySide2 import QtCore
p = QtCore.QProcess()
p.start("pointless")
p.waitForFinished(-1)
```

This gives us the first opportunity to improve things. You should modify your QProcess() calling code to
```python
from PySide2 import QtCore
p = QtCore.QProcess()
p.setStandardInputFile(QtCore.QProcess.nullDevice())
p.start("pointless")
p.waitForFinished(-1)
```
when you [know you do not want to pass anything into `stdin`](https://doc.qt.io/archives/qtforpython-5.12/PySide2/QtCore/QProcess.html#PySide2.QtCore.PySide2.QtCore.QProcess.setStandardInputFile), and/or
```python
from PySide2 import QtCore
p = QtCore.QProcess()
p.start("pointless")
p.closeWriteChannel()
p.waitForFinished(-1)
```
when you do [want to write, but know you're done](https://doc.qt.io/archives/qtforpython-5.12/PySide2/QtCore/QProcess.html#PySide2.QtCore.PySide2.QtCore.QProcess.closeWriteChannel).

Because `subprocess.run` (same as `procrunner.run`) by default forwards the incoming `stdin` this means
```python
from PySide2 import QtCore
p = QtCore.QProcess()
p.start("python3", ["-c", "import subprocess;subprocess.run(['pointless'])"])
p.waitForFinished(-1)
```
will also hang, whereas
```python
from PySide2 import QtCore
p = QtCore.QProcess()
p.start("python3", ["-c", "import subprocess;subprocess.run(['pointless'], stdin=subprocess.DEVNULL)"])
p.waitForFinished(-1)
```
will not.
Now here is an interesting question - why does this not hang?
```python
import subprocess
subprocess.run(["pointless"])
```
Because that too forwards an open (unclosed, blocking readable yet empty) stdin to `pointless`. (And that's when you hit the point of _'How could this ever have worked?'_)

### Enter `strace`.
When run via QProcess pointless does:
```
open("/var/lib/sss/mc/passwd", O_RDONLY|O_CLOEXEC) = 4
fstat(4, {st_mode=S_IFREG|0664, st_size=8825056, ...}) = 0
mmap(NULL, 8825056, PROT_READ, MAP_SHARED, 4, 0) = 0x7f4ee6c73000
fstat(4, {st_mode=S_IFREG|0664, st_size=8825056, ...}) = 0
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853261902
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853261902
ioctl(0, TCGETS, 0x7ffd188d87c0)        = -1 ENOTTY (Inappropriate ioctl for device)
write(1, "Ignoring unrecognised switch \"--"..., 644) = 644
fstat(0, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f4eea1fb000
read(0,  <unfinished ...>)              = ?
+++ killed by SIGKILL +++
```
and hangs. When run directly (with output redirected to /dev/null):
```
open("/var/lib/sss/mc/passwd", O_RDONLY|O_CLOEXEC) = 4
fstat(4, {st_mode=S_IFREG|0664, st_size=8825056, ...}) = 0
mmap(NULL, 8825056, PROT_READ, MAP_SHARED, 4, 0) = 0x7f16bec22000
fstat(4, {st_mode=S_IFREG|0664, st_size=8825056, ...}) = 0
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
ioctl(0, TCGETS, {B38400 opost isig icanon echo ...}) = 0
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
write(1, "Ignoring unrecognised switch \"--"..., 676) = 676
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
write(1, "\n**** ERROR ****\n", 17)     = 17
write(2, "No HKLIN filename given", 23) = 23
write(2, "\n", 1)                       = 1
futex(0x7f16c217d1d0, FUTEX_WAKE_PRIVATE, 2147483647) = 0
times({tms_utime=1, tms_stime=1, tms_cutime=0, tms_cstime=0}) = 853339586
write(1, "\n$TEXT:Reference: $$ Please cite"..., 442) = 442
exit_group(1)                           = ?
+++ exited with 1 +++
```
File 0 obviously is `stdin`. In a QProcess environment pointless tries to call [`ioctl(0, TCGETS)`](https://linux.die.net/man/2/ioctl) and fails, presumably assumes it is therefore reading from a non-terminal, does an `fstat` and then reads from the device, forever. Whereas in the semi-interactive environment (where it could read from the terminal but doesn't) it gets the terminal information, derives that this is a terminal, does not attempt an `fstat`, and does not attempt to read from the terminal.
This is commensurate with `pointless` behaviour when run directly - it will not wait on `stdin`.

And indeed you can easily reproduce a hanging pointless with
```
mkfifo garbage
cat > garbage & strace pointless < garbage
```
I don't think pointless is really to blame here though. It is just waiting on a file which for all intents and purposes looks like a legitimate input file.

That leaves us with two more options. We can in procrunner set `stdin=subprocess.DEVNULL` instead of the default `None` (ie. `stdin` forward), however I think this on balance has more downsides than upsides. Especially since the intent of procrunner is that most use cases of `procrunner.run()` will eventually be replaced by `subprocess.run()`, which also has a `stdin=None` default (presumably for $reasons), so this would just come back to bite us another day.

Finally, we can decide to not forward `stdin` in xia2. This makes total sense, and we should probably do this everywhere.

PR #652, opened 2022-01-28
Add environment variable to force installation in build/

ndevenish:
If the `CCTBX_INSTALL_PACKAGE_BUILD` environment variable is set, this bypasses the logic that checks to see if the base directory is writable and installs directly into the build folder.


libtbx_refresh.py
```diff
@@ -136,7 +136,11 @@ def _show_xia2_version():
 #
 # We need to check before trying to install as pip does os.access-based
 # checks then installs with --user if it fails. We don't want that.
-if _test_writable_dir(Path(site.getsitepackages()[0])):
+# Additionally, if the CCTBX_INSTALL_PACKAGE_BUILD environment variable
+# is set, then we explicitly want to do the readonly fallback.
+if ("CCTBX_INSTALL_PACKAGE_BUILD" not in os.environ) and _test_writable_dir(
```
Anthchirp, line 141:
```suggestion
if not os.getenv("CCTBX_INSTALL_PACKAGE_BUILD") and _test_writable_dir(
```
allows unsetting the variable using `set CCTBX_INSTALL_PACKAGE_BUILD=`

Anthchirp (approved):

PR #655, opened 2022-02-10
Add an updated setup.py

ndevenish:
This updates `setup.py` to:
- Have a PEP440 compatible version identifier (`X.Y.Z+<commit>-<branch>` or `X.Y.devN+<commit>-branch`)
- Fill out a few more fields
- Include package data when installing in non-editable mode

~The latter has required adding `__init__.py` to folders without it - so they are easily classed as a package. There is probably a way to do this without that change, but this was the easiest route. The folders with this added are:~
- `xia2/Data`
- `xia2/templates`
- `xia2/css`
- `xia2/Test/Handlers`, `/Modules` and `regression/expected`.

The latter of these - they probably don't need to be in the main xia2 source any more? It'd be nice to move them out to the `tests` dir if we can, because otherwise they _need_ to be included with the package install to successfully run the tests against a fully installed xia2.

Other TODO:
- [x] The `install_requires` list is from 2019 so I don't know how out of date it is
- [x] Decide on `xia2/Test`
- [x] Check the version parsing works how we want

When this change is made, it means that both `xia2` and `dxtbx` can be installed as proper packages, which should enable us to use conda-pack and solve the "readline doesn't work" and related issues.


ndevenish:
In developing the WIP `src` branch for DIALS, I worked out how to include package data without an `__init__.py` in every folder, so have removed those changes.

ndevenish:
Okay, so; 
- The `install_requires` list almost certainly isn't complete but the packages listed aren't wrong either, so probably no harm in leaving it be for now
- Deferring decision on testing for another day

It's just validating the version-specifier part? and review?

setup.py
```diff
@@ -48,26 +49,32 @@
...
-    if not os.path.exists(os.path.join(xia2_root_path, ".git")):
+    if not xia2_root_path.joinpath(".git").is_dir():
         return None
 
     try:
         result = subprocess.run(
-            ("git", "describe", "--long"),
+            ("git", "describe", "--tags", "--long", "--first-parent"),
             check=True,
-            cwd=xia2_root_path,
```
Anthchirp, line 60: I believe this should not be removed?

setup.py
```diff
@@ -78,22 +85,22 @@ def get_git_revision():
...
-        "tabulate",
-        'importlib_metadata;python_version<"3.8"',
-    ],
+setuptools.setup(
+    version=get_git_revision() or __version_tag__,
+    package_data={
+        "": ["*"],
+        "xia2": ["Data/*", "css/*", "templates/*"],
+        "xia2.Test": ["Handlers/*", "Modules/*", "regression/expected/*"],
+    },
```
Anthchirp, line 103: not sure this is necessary at all?
(or if it is: should likely be done differently)

Anthchirp (approved):

setup.py
```diff
@@ -1,9 +1,10 @@
...
-import os
+import re
 import subprocess
+from pathlib import Path
 
-from setuptools import setup
+import setuptools
 
 # Version number is determined either by git revision (which takes precendence)
 # or a static version number which is updated by bump2version
```
Anthchirp, line 10: should this be the other way round? Static version number takes precedence, and only if it's a dev version should we even attempt to get a git version number?

ndevenish: I think you are right here - I was attempting to handle an edge case, but by complicating and introducing more possibility for error in all other cases. Will rework this.

setup.py
ndevenish: Hmm, apparently pip explicitly moves the CWD to the installing package. At least we're more robust to other running with it explicitly set.

setup.py
ndevenish: I'm attempting to lean on the side of caution to fully replicate the current contents, ignoring the questions of whether they should be here. As such, unless you have strong objections or know the "proper" way we're supposed to do this, I'd prefer to keep this part temporarily intact?

ndevenish: Okay, after playing a bit, this does cover both `bdist` and `sdist`. I'm going to merge this in, but with the qualifier that if it's somewhat broken in another way, people can blame me and I will fix.

PR #659, opened 2022-02-21
Replace own which() function with shutil.which()

Anthchirp:


graeme-winter (approved):
No idea when `shutil.which` appeared but seems sensible to remove current implementation

Thank you

No idea why you found this though 🙃

Anthchirp:
Stumbled across `shutil.which` earlier, and remembered that we had a custom implementation somewhere. Couldn't see anything in dlstbx so suspected it was here :shrug:

PR #660, opened 2022-03-14
Add new experimental SSX processing pipeline dev.xia2.ssx

jbeilstenedmands:


jbeilstenedmands:
I am now splitting this work up into two pull requests, first for the data integration then a second for data reduction.

PR #664, opened 2022-04-06
Use smaller datasets in some regression tests

Anthchirp:
Instead of the relatively large 'insulin' dataset use
'centroid-test-data' for some regression tests.

This reduces test times between 50% and 75% for the modified tests,
and allows more tests to be enabled to run without `--regression-full`.

Two tests were enabled to run on Azure (one due to runtime reduction,
the other due to runtime reduction and the removal of an incorrect CCP4
dependency declaration)

### DIFF
--- a/tests/Modules/Indexer/test_DIALS_indexer.py
+++ b/tests/Modules/Indexer/test_DIALS_indexer.py
@@ -16,11 +16,10 @@ from xia2.Schema.XSweep import XSweep
 from xia2.Schema.XWavelength import XWavelength
 
 
-def exercise_dials_indexer(dials_data, tmp_path, nproc=None):
-    if nproc is not None:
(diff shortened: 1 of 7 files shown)


tests/Modules/Indexer/test_DIALS_indexer.py
```diff
@@ -68,10 +65,10 @@ def exercise_dials_indexer(dials_data, tmp_path, nproc=None):
     indexer2.eliminate()
 
     assert indexer.get_indexer_cell() == pytest.approx(indexer2.get_indexer_cell())
-    assert indexer.get_indexer_lattice() == "hR"
-    assert indexer2.get_indexer_lattice() == "hR"
+    assert indexer.get_indexer_lattice() == "oC"
```
graeme-winter, line 68: This data should be tP? 🤔

PR #666, opened 2022-04-08
Deprecate PHIL parameter 'report.resolution_bins'

### LINKED ISSUE #665: Synchronise n_bins across text output and report output
e.g. make `report.resolution_bins` pick up the same `n_bins` as merging, and deprecate the former. Confuses even `xia2` developers, even if it made sense at the time...

benjaminhwilliams:
Deprecate the PHIL parameter `xia2.settings.report.resolution_bins`, since it leads to confusion with `xia2.settings.merging_statistics.n_bins`.  The former is used for the HTML report and the latter for the `*_merging-statistics.txt` file in the `LogFiles` directory.  In future, we will simply use `n_bins` for both cases.

As it stands, I've marked `resolution_bins` for removal in version 3.10.

Requested @ndevenish's review because this relates to release cycles.  Have I chosen the correct deprecation timeline?
Requested @graeme-winter's review because I'm not confident I understand the preferred way to handle the xia2 PHIL parameters.  See comments below.

Closes #665.


src/xia2/Modules/Report/__init__.py
```diff
@@ -74,6 +78,16 @@ def __init__(
     ):
 
         self.params = params
+        self.n_bins = PhilIndex.params.xia2.settings.merging_statistics.n_bins
```
benjaminhwilliams, line 81: Two questions, perhaps for @graeme-winter:
1. Am I using `xia2.Handlers.Phil.PhilIndex` correctly, or should I be using its `get_python_object` method?
2. Until I introduced this line, `Report`'s parameters could all be passed in as an attribute at initialisation time, albeit as a cumbersom `phil.scope_extract` object.  Now I'm getting `PhilIndex` and I don't fully understand when its values get set.  Am I introducing any subtle weirdness?

ndevenish:
Merge pending @graeme-winter review

src/xia2/Modules/Report/__init__.py
graeme-winter: 1. use of `PhilIndex` looks fine
2. I don't know if you are adding subtle weirdness but your concern is valid - however I think if you have a `PhilIndex` then I think this is functionally the same as a `scope_extract`?

src/xia2/Modules/Report/__init__.py
```diff
@@ -74,6 +78,16 @@ def __init__(
...
+        # Remove this if statement block, and the corresponding PHIL parameter
+        # xia2.settings.report.resolution_bins, after version 3.9.
+        if self.params.resolution_bins:
```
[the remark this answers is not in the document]
graeme-winter: Once option is removed will an error occur or will it be silently ignored?

tests/test_report.py
```diff
@@ -1,27 +1,28 @@
...
+def report(dials_data, tmp_path_factory):
     data_dir = dials_data("pychef", pathlib=True)
     mtz = data_dir / "insulin_dials_scaled_unmerged.mtz"
-    temp_path = tmpdir_factory.mktemp("test_report")
-    with temp_path.as_cwd():
-        from xia2.Modules.Analysis import (  # import creates /xia2-debug.txt dropping
-            phil_scope,
-        )
-        from xia2.Modules.Report import Report  # import creates /DEFAULT/ dropping
+    tmp_path = tmp_path_factory.mktemp("test_report")
```
graeme-winter, line 16: Good bit of housekeeping 👍

tests/test_report.py
```diff
@@ -108,6 +110,65 @@ def test_resolution_plots_and_stats(report):
     }
 
 
+def test_deprecated_resolution_bins(dials_data, tmp_path, caplog):
```
graeme-winter, line 113: Glorious bit of overkill 🙂

graeme-winter (approved):
Change set looks sensible - this is very corner case anyway so I think it is relatively low risk. The concern about introducing subtle bugs is valid but I think should not stop this from going in. I also cannot tell if this will add subtle bugs either... which indicates that PHIL is unhelpful.

src/xia2/Modules/Report/__init__.py
benjaminhwilliams: OK, thanks.  I try to remain ignorant of the weird workings of PHIL, so if you don't have any concerns, nor do I.  It's true that `PhilIndex`'s type is a `scope_extract`, so it's more a question of when the scope is extracted (xia2 has lots of import-time initialisation of stuf like this) and whether that matters.

I think if you don;t have any concerns, we can probably merge it and find out the hard way.

PR #672, opened 2022-05-16
Bugfix for finding NeXus data

### LINKED ISSUE #671: processing multiple run datasets - nxs files from Tristan 10M detector
Hi, 

During the beamtime we could not process multi-run datasets using the normal routines.

visit: /dls/i19-2/data/2022/cy30210-1

datasets to process together: 
206a_PtCmpd_100K_50kHz_ND2_SS
206b_PtCmpd_100K_50kHz_ND2_SS

nxs files: 
/dls/i19-2/data/2022/cy30210-1/processing/206a_PtCmpd_100K_50kHz_ND2_SS/PtCmpd_1_images.nxs
/dls/i19-2/data/2022/cy30210-1/processing/206b_PtCmpd_100K_50kHz_ND2_SS/PtCmpd_2_images.nxs

example xia2 command: 
xia2 small_molecule=true image=/dls/i19-2/data/2022/cy30210-1/processing/206a_PtCmpd_100K_50kHz_ND2_SS/PtCmpd_1_images.nxs image=/dls/i19-2/data/2022/cy30210-1/processing/206b_PtCmpd_100K_50kHz_ND2_SS/PtCmpd_2_images.nxs

Please can you have look into why this might be?

graeme-winter:
If the data are not following the Eiger conventions but are otherwise correct they may be ignored (to be precise: second and subsequent sweeps may be ignored which is frustrating) - handle that by simply taking data as written if they are following the normal convention.

Fixes #671

The fix for this is not pretty and indicates that a proper refactor of this whole area would be appropriate

### DIFF
--- a/src/xia2/Applications/xia2setup.py
+++ b/src/xia2/Applications/xia2setup.py
@@ -247,6 +247,9 @@ def _filter_aliased_hdf5_sweeps(sweeps):
                 rest.append(s)
             continue
         filenames = tuple(_list_hdf5_data_files(s))
+        if not filenames and s not in rest:
+            rest.append(s)
+            continue
         if filenames in h5_data_to_sweep:
             # impose slight bias in favour of using _master.h5 in place of .nxs
             # because XDS
--- a/newsfragments/XXX.bugfix
+++ b/newsfragments/XXX.bugfix
@@ -1,0 +1,2 @@
+xia2 with NeXus files: if the NeXus files are not from an Eiger some are ignored, as the /entry/data/data_[nnnn] entries do not exist -> in this case accept the data as passed in rather than trying to do "clever workarounds"
+


benjaminhwilliams:
I'm not sure I fully understand the failure mode.  Do you have a little demo?

graeme-winter:
Failure mode was only one sweep was found when you passed in two for processing. You can identify this by running 

```
xia2setup /path/to/data/folder
```

where there are 2 runs - only one is found.

This is a sinister failure mode as no error is emitted, just half of your data are ignored.

rjgildea:
Please could you add some comments/docstrings to the `_filter_aliased_hdf5_sweeps()` function? It isn't immediately obvious to me what it is supposed be doing...

graeme-winter:
> Thanks, appreciated. If I understand correctly the main purpose of this function is to avoid finding the `prefix_master.h5` and `prefix.nxs` file twice (in the case of DLS datasets)? 👍

Precisely this

src/xia2/Applications/xia2setup.py
```diff
@@ -238,6 +238,24 @@ def _list_hdf5_data_files(h5_file):
...
+    - with non Eiger data, need to make sure that data are found correctly
+      (believe this is now resolved)
+    - results not well defined if you have multiple views on the data
+      deliberately e.g. when you have 14,400 image scan "split into four"
+      with four NeXus files pointing into 15 data files...
+    - proper use of /entry/data/data, as a VDS. with no data_000001...N may
+      result in this aliasing function missing alises e.g. if master.h5 and
+      nxs files exist pointing into the same image sequence both with VDS
+    """
+
```
rjgildea, line 258: Add note to self:
```suggestion
    """
    Filter HDF5 (or NeXus) data files which share the same underlying data.
    
    Avoiding the case where you have two NeXus / master files which represent
    the same experimental measurements, e.g. in the case of DLS data,
    where prefix_master.h5 and prefix.nxs point to the same underlying
    dataset. N.B. this is rather heavily tied to the HDF5 data representation
    which relies on /entry/data/data_000001...N individual data sets.

    Known issues:
    - with non Eiger data, need to make sure that data are found correctly
      (believe this is now resolved)
    - results not well defined if you have multiple views on the data
      deliberately e.g. when you have 14,400 image scan "split into four"
      with four NeXus files pointing into 15 data files...
    - proper use of /entry/data/data, as a VDS. with no data_000001...N may
      result in this aliasing function missing alises e.g. if master.h5 and
      nxs files exist pointing into the same image sequence both with VDS
    """
```

benjaminhwilliams:
> Thanks, appreciated. If I understand correctly the main purpose of this function is to avoid finding the `prefix_master.h5` and `prefix.nxs` file twice (in the case of DLS datasets)? +1

Ah, OK, I didn't understand that on the first few passes.

I'm still reviewing this and I'll have a few changes, but they're not yet ready.  Can I please beg a little more time for my review?

benjaminhwilliams:
> Thanks, appreciated. If I understand correctly the main purpose of this function is to avoid finding the `prefix_master.h5` and `prefix.nxs` file twice (in the case of DLS datasets)? +1

Ah, OK, I didn't understand that on the first few passes.

I'm still reviewing this and I'll have a few changes, but they're not yet ready.  Can I please beg a little more time for my review?

graeme-winter:
Of course you can have a little more time. If you have comments or questions there is almost certainly more to do here 👍

benjaminhwilliams:
@graeme-winter, does this still behave as intended, or have I missed something/messed something up?

benjaminhwilliams:
It may be valuable, now that I understand the intent, to add some tests of the various possible inputs, and perhaps even some tests to demonstrate the known limitations.

src/xia2/Applications/xia2setup.py
```diff
@@ -226,46 +226,87 @@ def visit(directory, files):
...
-            continue
-        filenames = tuple(_list_hdf5_data_files(s))
-        if filenames in h5_data_to_sweep:
-            # impose slight bias in favour of using _master.h5 in place of .nxs
-            # because XDS
-            if h5_data_to_sweep[filenames].endswith(".nxs") and s.endswith(
-                "_master.h5"
-            ):
-                h5_data_to_sweep[filenames] = s
+        if not is_hdf5_name(s) or not (filenames := _linked_hdf5_data_files(s)):
```
graeme-winter, line 293: @benjaminhwilliams  superb if sneaky use of walrus operator _however_ ISTR such are verboten until our chums at CCP4 have a platform which supports same? Idiomatically excellent application of same though.

src/xia2/Applications/xia2setup.py
```diff
@@ -226,46 +226,87 @@ def visit(directory, files):
...
+    data_path = "/entry/data"
+    with h5py.File(h5_file) as f:
+        filenames = [
+            f[data_path][k].file.filename for k in f[data_path] if k.startswith("data_")
+        ]
+    return frozenset(filenames)
+
+
+def _filter_aliased_hdf5_sweeps(sweeps: list[str]) -> set[str]:
+    """
```
graeme-winter, line 239: Kudos and respect for the excellent docstring

graeme-winter (commented):
OK, read through the change set for understanding rather than execution (I assume that the subsequent commits do not alter behaviour) 

Excellent improvement to the code base. Thank you. I cannot approve your changes as this is my PR. Sneaky walrus should be removed though.

src/xia2/Applications/xia2setup.py
benjaminhwilliams: Ah, I'd forgotten that walruses are forbidden.  Since CI checks run on 3.8 & 3.9 only, I assumed all distribution was OK.  Are CCP4 really still trying to release on Python < 3.8?

PR #674, opened 2022-05-20
Slightly less stupid MTZ rename

### LINKED ISSUE #673: If you run xia2 in a directory which contains the letters "mtz" you will not go to space that day
It turns out there is a dumb replace in some of the filename code which replaces "mtz" with "sca" which obviously works 100% of the time unless you have "mtz" somewhere in your file path

🤦‍♂️

Reported by @dagewa

graeme-winter:
Fixes #673

Propose this as a "quick fix" which we should merge quickly while we implement a slightly better fix for everywhere where this kind of string replacement is performed. I had a quick grep through and a proper fix is likely to be a nontrivial undertaking.


dagewa (approved):
Just happened to me again!

PR #684, opened 2022-06-15
Replace procrunner use with subprocess

rjgildea:
Given procrunner was intended primarily to provide backwards compatibility to legacy Python versions (which we no longer support), it makes sense to use the standard Python library `subprocess.run()` method in its place.

Full regression tests pass with:
```
$ pytest --regression-full -n auto -vs
```


Anthchirp:
As author of procrunner I can state that this is a welcome move :slightly_smiling_face:

PR #700, opened 2022-08-18
Multiplex set output filenames

graeme-winter:
Cleaning up some useless code in multiplex and adding one useful feature. Will work on other issue in a new branch.


graeme-winter:
Converting to draft while I add another feature

jbeilstenedmands:
Hi, just looking to tidy up the PR list and propose we close this one, but feel free to reopen if you really want.

I note that since you made this, it was also brought up as an issue in dials:
https://github.com/dials/dials/issues/3052
So I propose that once that work is done, we can use the same idea for xia2.

PR #733, opened 2023-05-02
Ssx clustering

amyjaynethompson:
Separated out the clustering algorithm from multiplex and made it applicable to SSX data (as well as still accepting 'standard' rotation multi-crystal datasets). Some minor edits to the original clustering methods to work around the fact that the .expt file in SSX data has an empty 'scan' field. An initial algorithm to try and pick out 'interesting' clusters is included, but would greatly benefit from feedback and some pointers for improvement prior to branch merging.


src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    reflections_all = flex.reflection_table()
+    assert len(reflections) == 1 or len(reflections) == len(experiments)
+    for i, (expt, refl) in enumerate(zip(experiments, reflections)):
+        reflections_all.extend(refl)
+    reflections_all.assert_experiment_identifiers_are_consistent(experiments)
+
+    try:
+        MCA = MultiCrystalReport(params_mca, experiments, reflections_all)
+
```
jbeilstenedmands, line 135: Very excited to see this functionality coming to xia2 😃 . I'm just starting to test now, and I got the following error at this line:
```  
File "/dials/modules/xia2/src/xia2/cli/cluster_analysis.py", line 141, in run
    MCA._cluster_analysis.cc_clusters
AttributeError: 'NoneType' object has no attribute 'cc_clusters'
```
I think the clustering was not being run, is this line missing?
```suggestion
MCA.cluster_analysis()
```
Also, in order to run the program, it looks like you need to supply a phil file with at least the parameters `max_cluster_height_difference`,`min_cluster_size`, `max_output_clusters`. The standard way we do this in the programs is to have all options defined in the phil scope with suitable default values and a nice description, so that it is not necessary to provide a specific phil file as input. Then, if a phil file is provided, this is a mechanism to overwrite the default values.

amyjaynethompson: Yep! Unsure how the missing line happened - but I've put it back in :) 

Also confused myself with pushing separate branches and forgot to add the phil params - I put them in ScaleAndMerge.py since that's where most of the multiplex parameters are and this particular function of finding interesting clusters could eventually find use in multiplex as well - should be in this thread now :)

jbeilstenedmands (changes requested):
Hi Amy,

Here is my overall feedback/suggestions for changes which I think will help make this a great addition to xia2. Thanks again for contributing this! 🙂 

1)	Changes to the phil scope – I put a suggestion on the pull request, but it will need a few corresponding changes in the run function. The main motivation is to make sure only the relevant parameters are shown for the program’s options, rather than the full set of multiplex parameters.
2)	Use of the `ssx_flag` option. The ideal functionality here is that the program should do something sensible whether the input is rotation or serial data, so that it is not necessary to have to set `ssx_flag=True/False` as an input option. We will still need to have an internal parameter in the `DataManager`, I put a suggestion in the pull request as to how to assess whether the data are stills or sweeps. An internal flag should reduce some of the branching in the code where you pass the `ssx_flag` as arguments to the class methods, but there will still need to be some if/else statements when handling things like the batches.
3)	Logging. A little log output would help – even if this is just making sure the output in the `.txt` files is also printed to the log output and an indication that data has been saved into the cluster directories.
4)	Testing. A test will be important to make sure the program keeps working long into the future. I think the minimum would be a ‘regression’ test in e.g. `tests/regression/test_cluster_analysis.py` You can look in `tests/regression/test_multiplex.py` or `test_ssx.py` for example of how this can be done. There are some datasets available aleady that you could use - `multi_crystal_proteinase_k` are some processed narrow rotation datasets and there are some processed serial datasets in `cunir_serial_processed` and `dtpb_serial_processed`. These examples only have a few images/sweeps in, so they don’t really represent a real-life use case unfortunately. There are a few things we could do to address that, but I would consider these optional and not a blocker for getting this merged (perhaps something for in future). a) - Use some of the data you have been using to test the program. This would require putting the processed data somewhere publicly available (e.g. `https://github.com/dials/data-files`), the main concerns here are not having a file size that is too big and also not having a test that takes too long to run b) – Write an extra functional test for just the `interesting_cluster_identification` function. I think this function could easily be refactored to be a `staticmethod` of the class – then in a test you could create lists of `ClusterInfo` objects and test the behaviour of your function for different parameter options. This will be a much quicker test to run as you avoid having to run the clustering algorithm each time, and you can easily test more input combinations.
5)	Documentation. This is something we have not been good at historically, so don’t consider this as mandatory, but I feel that lots of people will want to try this program, so being able to point them to something on the webpage will help make this more impactful. It’s not clear to me at the moment where would be the best place for this to go (e.g. there is an in depth multiplex tutorial at https://dials.github.io/documentation/tutorials/br_lyso_multi.html , and the xia2 webpage only has documentation for the full pipelines and not the utility programs https://xia2.github.io/). I don’t think a full tutorial is necessary, I was thinking along the lines of what I wrote for xia2.ssx: https://xia2.github.io/serial_crystallography.html# i.e. a description of how to run the program, what are the main options to be aware of, a short description of what the algorithm is doing and what the output of the program is.

src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,240 @@
...
+  .short_caption = "Minimum number of datasets for an important cluster"
+output_correlation_cluster_number = 0
+  .type = int
+  .short_caption = "Option to output a specific correlation cluster when re-running the code"
+output_cos_cluster_number = 0
+  .type = int
+  .short_caption = "Option to output a specific cos cluster when re-running the code"
+
+output {
+  log = xia2.multi_crystal_analysis.log
```
jbeilstenedmands, line 64: Should this be `log = xia2.cluster_analysis.log`?

src/xia2/Modules/MultiCrystal/data_manager.py
```diff
@@ -27,22 +27,44 @@ def __init__(self, experiments, reflections):
...
+                    # This may be a different scan instance ¯\_(ツ)_/¯
+                    expt.imageset.get_scan().set_batch_offset(
+                        expt.scan.get_batch_offset()
+                    )
+                logger.debug(
+                    f"{expt.scan.get_batch_offset()} {expt.scan.get_batch_range()}"
+                )
+
+            else:
+                self.ssx_batch_dict[expt] = i * 10**n
```
jbeilstenedmands, line 66: I don't think the ssx_batch_dict is used for anything? So we can remove it.

src/xia2/Modules/MultiCrystal/data_manager.py
```diff
@@ -27,22 +27,44 @@ def __init__(self, experiments, reflections):
...
     def _set_batches(self):
-        max_batches = max(e.scan.get_image_range()[1] for e in self._experiments)
-        max_batches += 10  # allow some head room
+
+        if not self.all_stills:
+            max_batches = max(e.scan.get_image_range()[1] for e in self._experiments)
+            max_batches += 10  # allow some head room
+        else:
+            max_batches = 1
+            self.ssx_batch_dict = {}
```
jbeilstenedmands, line 48: I don't think the ssx_batch_dict is used for anything? So we can remove it.

src/xia2/cli/cluster_analysis.py
amyjaynethompson: Yes it should be - this has been changed

src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,240 @@
...
+            cos_angle_cluster_table=MCA._cos_angle_cluster_table,
+            cos_angle_cluster_json=MCA._cos_angle_cluster_json,
+            image_range_tables=[table_list],
+            xia2_version=Version,
+        )
+
+        with open("xia2.cluster_analysis.html", "wb") as f:
+            f.write(html.encode("utf-8", "xmlcharrefreplace"))
+
+        print("Clusters recommended for comparison in xia2.cluster_analysis.log")
```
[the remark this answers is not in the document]
amyjaynethompson: Yep, am pointing to the log file, but have changed to a logger.info rather than a print

src/xia2/Modules/MultiCrystal/data_manager.py
amyjaynethompson: Removed :)

src/xia2/Modules/MultiCrystal/data_manager.py
amyjaynethompson: Also removed :)

src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,284 @@
...
+            table_list.append([id_list[j], i.paths()[0]])
+
+        loader = ChoiceLoader(
+            [PackageLoader("xia2", "templates"), PackageLoader("dials", "templates")]
+        )
+        env = Environment(loader=loader)
+
+        template = env.get_template("clusters.html")
+        html = template.render(
+            page_title=params.title,
```
jbeilstenedmands, line 259:
```suggestion
            page_title="xia2 cluster analysis",
```
If we set the title here, then we can remove a few phil options.

src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,284 @@
...
+
+logger = logging.getLogger("xia2.cluster_analysis")
+
+mca_phil = iotbx.phil.parse(
+    """
+title = 'xia2 report'
+    .type = str
+prefix = 'xia2'
+    .type = str
+
```
jbeilstenedmands, line 34:
```suggestion
```
With the change below, we don't need the title, and I don't see that the prefix is used

src/xia2/cli/cluster_analysis.py
```diff
@@ -0,0 +1,284 @@
...
+seed = 42
+  .type = int(value_min=0)
```
jbeilstenedmands, line 36:
```suggestion
  .type = int(value_min=0)
  .help = "Seed value for random number generators used"
```

jbeilstenedmands (approved):
Once the last few points are addressed then I think this is ready to be merged.
I will merge in the main branch again later this afternoon to make sure all the tests are passing.
I know it was suggested that this kind of program should be in dials rather than xia2, however the code is closely tied into the multiplex code and I don't see that it can be cleanly extracted. If we wanted to move it, then parts of the multiplex code would also need to be moved or refactored, so this is something for another time and should not delay this work being merged.
I tested on 1.6K still images and it ran fine. Running on 16K images is proving much slower, but that is due to the performance of `dials.cosym` which is a known issue for these extremes.
Thanks again!

PR #734, opened 2023-05-02
Multiplex ref

amyjaynethompson:
Added in reference parameter for multiplex to be passed to cosym. Also added a parameter to tailor the cluster output - option to output a .csv file with the datasets included in each cluster rather than a .mtz to save on computing time. Output .csv can be easily input into other code - but feedback is welcome if there is a better way of doing this.


jbeilstenedmands (changes requested):
I think this pull request is introducing two concepts, the use of a reference and changing the output clustering. It would be best to just focus on the changes needed for supporting use of a reference in this pull request, and make a separate pull request for the clustering changes.

src/xia2/Wrappers/Dials/Cosym.py
```diff
@@ -67,6 +68,9 @@ def set_best_monoclinic_beta(self, best_monoclinic_beta):
         def set_lattice_symmetry_max_delta(self, lattice_symmetry_max_delta):
             self._lattice_symmetry_max_delta = lattice_symmetry_max_delta
 
+        def set_reference_pdb(self, reference_pdb):
```
jbeilstenedmands, line 71: I think it would be clearer to call this property and associated methods `reference_file` rather than `reference_pdb`, as we support pdb, mtz, cif...

amyjaynethompson: Done :) Will be located on the separated branch.

amyjaynethompson:
Have separated out these two new features. Let me know if you have any other suggestions.

jbeilstenedmands:
Now that https://github.com/dials/dials/pull/2420 is merged, I think what we need to do is slightly different. Rather than using a reference in cosym, which doesn't seem to be robust, we should call dials.reindex after the scaling and symmetry analysis, which i think is at this point:
https://github.com/xia2/xia2/blob/fe693e89d1c737809b6f1e677a77466c37c40e24/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py#L368

After this point there is no further symemtry analysis, so all clusters etc should be in the same setting.

PR #744, opened 2023-06-09
Cluster update

amyjaynethompson:
Updates to the algorithm to select potentially interesting clusters - it is more reliable now in what it selects based on current research directions.


ndevenish:
Any easy summary of what has actually changed in the algorithm?

amyjaynethompson:
> Any easy summary of what has actually changed in the algorithm?

Both versions were more or less looking for the same thing - pairs of clusters with no data sets in common that might be interesting to compare. The original version was overly complex and consisted of "band-aid" solutions to cover edge cases - this new version was greatly simplified and streamlined by searching for the existence of a parent cluster that is the exact addition of two child clusters (with no datasets in common), which is what the current research is comparing. No band-aids required :)

jbeilstenedmands (approved):
Looks less complicated and seems to give similar results (looks like the only difference is from better filtering out of small clusters) 👍

PR #747, opened 2023-06-16
extend R-free flags for filtered.mtz to fix issue where incompatible

amyjaynethompson:
Came across a few mutliplex jobs where the filtered.mtz came out to a marginally higher resolution than the scaled.mtz. With the current code, this caused an issue where R-free flags were incompatible and thus could not be used for model refinement. I've set the r-free flags to be able to extend if needed while scaling the filtered data. Seems to have solved the problem for my problematic test cases.


jbeilstenedmands:
I'm still planning to review this before merging as I'm not sure the current change is suitable in general

ndevenish:
!? apparently auto-merge now merges _before_ tests have run.

PR #748, opened 2023-06-16
Multiplex reindex to ref

amyjaynethompson:
Added in the functionality for multiplex to run dials.reindex based on a reference file. After many tests and discussions it seemed to work better comparing to the reference PDB at this stage rather than at the initial cosym stage. Also edited the xia2 reindex wrapper for dials after the functionality for having a reference file (ie PDB) was added.

### DIFF
--- a/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
+++ b/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
@@ -28,6 +28,7 @@ from xia2.Modules.Scaler.DialsScaler import (
 from xia2.Wrappers.Dials.Cosym import DialsCosym
 from xia2.Wrappers.Dials.EstimateResolution import EstimateResolution
 from xia2.Wrappers.Dials.Refine import Refine
+from xia2.Wrappers.Dials.Reindex import Reindex
 from xia2.Wrappers.Dials.Scale import DialsScale
 from xia2.Wrappers.Dials.Symmetry import DialsSymmetry
 from xia2.Wrappers.Dials.TwoThetaRefine import TwoThetaRefine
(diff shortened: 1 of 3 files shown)


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -368,6 +379,8 @@ def __init__(self, experiments, reflections, params):
 
         self.decide_space_group()
 
+        self.reindex()
```
jbeilstenedmands, line 382: We should only run this `if params.reference is not None`,

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -771,6 +784,31 @@ def cosym(self):
...
+        logger.info("Re-indexing to reference")
+        experiments_filename = self._data_manager.export_experiments("tmp.expt")
+        reflections_filename = self._data_manager.export_reflections("tmp.refl")
+        reindex = Reindex()
+        auto_logfiler(reindex)
+        reindex.set_experiments_filename(experiments_filename)
+        reindex.set_indexed_filename(reflections_filename)
+        reindex.set_reference_file(self._params.reference)
+
+        if self._params.symmetry.space_group is not None:
```
jbeilstenedmands, line 798: We need to think carefully about what we want to happen if we set a reference file. Reindexing against a reference only make sense if the space groups are the same. So if we supply a reference file, then I think we should extract the space group from the reference file and set this as the space group at the start of multiplex processing (and check it doesn't clash with any user specified space group). Then we won't need to specify it here.

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
amyjaynethompson: Added this :)

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
amyjaynethompson: Think this should be updated as well - let me know if you think of a better way of doing this or anything else to add.

PR #758, opened 2023-08-02
Restructure xia2.ssx_reduce workflow

jbeilstenedmands:
This PR changes the xia2.ssx_reduce workflow to address a few issues:

Improved indexing ambiguity resolution - an error in the previous batch-based approach is fixed.
Size of processing on disk - by default now only save the scaled data files, not intermediate files (prefiltered, reindexed etc).
Reindexing to the correct hand when using a reference intensity set (following https://github.com/xia2/xia2/pull/748)
Avoiding one big scaling job if no reference model, by using a scaling-in-batches approach.
Reduce lattice_symmetry_max_delta from 2.0 to 0.5, to avoid catching cases where cell parameters are close but distinguishable and reindexing analysis can be detrimental.

Workflow for data with indexing ambguity:
1: cosym & scale on each batch,
2: cosym & scale batches together
3: If using a reference, reindex to reference and scale against reference

Workflow for data with no indexing ambiguity:
If have a reference intensity set, scale each batch against reference
If no reference, scale each batch, then scale batches together


jbeilstenedmands:
Closing in favour of a set of smaller changes addressing specific issues

PR #759, opened 2023-08-10
extend R-free flags for filtered.mtz to fix issue where incompatible

ndevenish:
Recreation of @amyjaynethompson's original https://github.com/xia2/xia2/pull/747, accidentally merged before ready

Came across a few mutliplex jobs where the filtered.mtz came out to a marginally higher resolution than the scaled.mtz. With the current code, this caused an issue where R-free flags were incompatible and thus could not be used for model refinement. I've set the r-free flags to be able to extend if needed while scaling the filtered data. Seems to have solved the problem for my problematic test cases.


jbeilstenedmands:
Depends on https://github.com/dials/dials/pull/2491

I have made a couple of changes here.

I have made the extend parameter an 'auto' type, which defaults to True (rather than forcing this parameter to always be True which is what the original changeset did).
We also need to consider what will happen for multiple clusters and/or filtering, and the potential for an external reference. I added some logic so that once the freer flags are extended to the full resolution range and completeness, this file is then used as the reference for all remaining merged mtz files, so that the sets are consistent across all files.

I would like @amyjaynethompson to test that everything still works as expected on the test dataset that motivated this PR, thanks!

amyjaynethompson (approved):
Fantastic! The updated logic still works for the test dataset. All output .mtz files (including clusters) refined successfully with the Rfree flags generated by this branch of multiplex. Thanks for your help in getting this issue fixed :)

PR #762, opened 2023-08-31
Add option to filter on mean I/sigma in xia2.ssx_reduce

jbeilstenedmands:
Via parameter `mean_i_over_sigma_threshold`


ndevenish:
@jbeilstenedmands do you recall if there was a reason this never made it in?

jbeilstenedmands:
No I think I just forgot about it, still looks good and it improved data quality when I was testing just now, so I think it is still a useful addition. Can be merged before or after release depending on your preference.

Issue #751, opened 2023-07-06
Multiplex: how does one use the `identifiers` option?


dagewa:
I would like to select the cluster merged by the purple line (on the upper dendrogram), but I failed to set `max_clusters` high enough when I ran the job. 
![newplot (4)](https://github.com/xia2/xia2/assets/11502083/4e258fa0-c8de-47f4-9c0f-8c9694f0902b)

I saw the `identifiers` option and figured perhaps it is a way to manually define a cluster of interest (I can't be sure as it has no PHIL help string). With that in mind, I copied the relevant datasets line out of the HTML table and re-ran my job like this, but it crashed:

```
$ xia2.multiplex ../scale-larger-cell-05/combined.{expt,refl} resolution.d_min=0.52 multi_crystal_analysis.unit_cell=14.467,21.862,17.717,90,90.11,90 identifiers="0 3 4 5 7 8 9 10 11 12 13 14 15 17 18 19 23 26 28 30 31 32 33 34 35 38 39 41 42 43 44 45 46 50 51 52 55 56 57 58 60 62 63 64 65 66"
DIALS 3.dev.997-gda59c1d80
The following parameters have been modified:

resolution {
  d_min = 0.52
}
multi_crystal_analysis {
  unit_cell = 14.467 21.862 17.717 90 90.11 90
}
identifiers = "0" "3" "4" "5" "7" "8" "9" "10" "11" "12" "13" "14" "15" "17" \
              "18" "19" "23" "26" "28" "30" "31" "32" "33" "34" "35" "38" \
              "39" "41" "42" "43" "44" "45" "46" "50" "51" "52" "55" "56" \
              "57" "58" "60" "62" "63" "64" "65" "66"
input {
  experiments = ../scale-larger-cell-05/combined.expt
  reflections = ../scale-larger-cell-05/combined.refl
}

/home/fcx32934/sw/cctbx/conda_base/lib/python3.10/site-packages/numpy/core/fromnumeric.py:3464: RuntimeWarning: Mean of empty slice.
  return _methods._mean(a, axis=axis, dtype=dtype,
/home/fcx32934/sw/cctbx/conda_base/lib/python3.10/site-packages/numpy/core/_methods.py:192: RuntimeWarning: invalid value encountered in scalar divide
  ret = ret.dtype.type(ret / rcount)
[...]

PR #764, opened 2023-10-03
Cluster bug fix

amyjaynethompson:
Found a bug in the cluster selection algorithm. Fix applied - the order in lists were inconsistent. Also added a line of documentation to address Issue #751


src/xia2/cli/cluster_analysis.py
```diff
@@ -268,6 +268,11 @@ def run(args=sys.argv[1:]):
             f.write(html.encode("utf-8", "xmlcharrefreplace"))
 
         logger.info(f"Clusters recommended for comparison in {params.output.log}")
+        logger.info("----------------")
+        logger.info("Output given as DIALS .expt/.refl files:")
+        logger.info("To merge rotation data: use dials.merge")
+        logger.info("To merge still data: use xia2.ssx_reduce")
```
jbeilstenedmands, line 274:
```suggestion
        logger.info("To merge still data: use xia2.ssx_reduce with the option steps=merge")
```
I think you are suggesting that the data from each cluster should be merged without further processing? If so it would be helpful to add this bit of text to avoid accidental rescaling of individual clusters.

jbeilstenedmands (approved):
Looks good, thanks! I have just made a suggested improvement for the log output.

PR #767, opened 2023-10-12
Add run_cluster_identification=True/False option to xia2.cluster_analysis

jbeilstenedmands:
Feature request, to allow people to run the program to just get the clustering html report without also doing the cluster identification analysis.


amyjaynethompson (approved):
Looks great! A very useful addition to save computation time where full analysis not needed and I had no errors when running it :)

PR #770, opened 2023-12-07
Avoid possibility of inconsistent space group settings if using a reference for multiplex

jbeilstenedmands:
@amyjaynethompson identified a case where multiplex results in an inconsistent space group setting for insulin when using a reference dataset. The multiplex data manager does extra reindexing as part of the space group decision, however we set the space group when using a reference, so we can skip this step which fixes the issue.


amyjaynethompson (approved):
Also required space group to be defined for dials.reindex - otherwise still got an inconsistent space group setting - now works for me with my added commit :) Thanks for all your help troubleshooting!

PR #771, opened 2023-12-20
Updates to handle indexed experiment lists containing experiments without crystals

jbeilstenedmands:
Depends on https://github.com/dials/dials/pull/2567


jbeilstenedmands:
Closing as this for now as is not currently planned, can be resurrected if needs to.

PR #778, opened 2024-02-21
Filter pandas Pyarrow deprecation warning in tests

jbeilstenedmands:
The latest version of pandas now includes a deprecation warning if Pyarrow is not installed, due to planned upcoming changes in pandas 3.0 (see https://github.com/pandas-dev/pandas/issues/54466)

Filter this warning in pytest for now to avoid test failures.


jbeilstenedmands:
This warning has now been suppressed as of pandas 2.2.1, so this PR is not needed

PR #783, opened 2024-03-20
Multiplex selected clusters

amyjaynethompson:
Re-vamping clustering behaviour within xia2.multiplex (and xia2.cluster_analysis). 

List of key changes:
- turn on clustering behaviour by setting `clustering.output_clusters=True` (originally needed to set `min_completeness` or `min_multiplicity` which some users have found to be unintuitive)
- can output cos-angle as well as correlation clusters in a single execution of the code (previously limited to one type per run of multiplex)
- set `clustering.max_cluster_height` to select a single cluster for output (ie can set `clustering.max_cluster_height` to the height of the desired cluster and `clustering.max_output_clusters=1` to limit the output to a single cluster) 
- integration of "interesting cluster identification" algorithm implemented within xia2.cluster_analysis - this identifies any pairs of clusters without any common datasets which may be interesting to compare and outputs the scaled/merged files (and follows behaviour of other clusters in the summary .html file)
- updated xia2.cluster_analysis and corresponding tests to be consistent with these new phil parameter definitions. 
- removed height comparison from xia2.cluster_analysis and interesting cluster identification algorithm as it wasn't really found to be helpful and has now been superseded by clustering.max_cluster_height


src/xia2/cli/cluster_analysis.py
```diff
@@ -40,32 +40,37 @@
...
-output_cos_cluster_number = 0
-  .type = int
-  .short_caption = "Option to output a specific cos cluster when re-running the code"
-exclude_correlation_cluster_number = 0
-  .type = int
-  .short_caption = "Option to output all data excluding a specific correlation cluster"
-exclude_cos_cluster_number = 0
-  .type = int
-  .short_caption = "option to output all data excluding a specific cos cluster"
+clustering
```
jbeilstenedmands, line 43: The changes to the options are good in general, I think multiplex and cluster_analysis should include a shared phil scope for these options, that is defined only in once place, so that the behaviour is the same for both. That way the clustering achieved by setting particular options in multiplex could be replicated by running xia2.cluster_analysis on the scaled.expt and scaled.refl from multiplex.

src/xia2/cli/multiplex.py
```diff
@@ -204,6 +198,30 @@ def run(args=sys.argv[1:]):
                 f"symmetry.space_group has been set to: {params.symmetry.space_group}"
             )
 
+    if params.clustering.output_clusters and not params.reference:
+        logger.info(
+            "WARNING: clustering selected but no reference given. "
```
jbeilstenedmands, line 203: Is this only an issue in the case of potential indexing ambiguity? Also I can't remember why this should occur anyway - is it because the clustering is performed on the scaled output - should we just be running on the reindexed file rather than scaled file and then this would solve this?

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -224,20 +224,48 @@
...
+  max_cluster_height = 100
+    .type = float
+    .short_caption = "Maximum height in dendrogram for clusters"
+  max_cluster_height_cc = 100
+    .type = float
+    .short_caption = "Maximum height in correlation dendrogram for clusters"
+  max_cluster_height_cos = 100
+    .type = float
+    .short_caption = "Maximum height in cos angle dendrogram for clusters"
+  analysis = False
```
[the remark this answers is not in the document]
amyjaynethompson: Have changed this parameter to `find_distinct_clusters` - while a bit wordy, it's no more wordy than some of the other parameters and gives a better idea as to what it does

src/xia2/cli/cluster_analysis.py
amyjaynethompson: Have re-worked this to put the clustering phil scope only in xia2.cluster_analysis - ScaleAndMerge then imports this. Also re-worked xia2.cluster_analysis to function more like multiplex.

src/xia2/cli/multiplex.py
amyjaynethompson: I'm not 100% sure - I mostly just remember that without a reference I would get clusters output from a single job of multiplex that were in different settings. For example, if I had 10 clusters output from a single job, only say 4 of them would refine successfully (this was again with cubic insulin). I'm not sure if we ever got to the bottom of why it was inconsistent, but I can look into this a bit further.

doc/sphinx/multiplex-multi-crystal.rst
```diff
@@ -0,0 +1,94 @@
...
+refinement, scaling, resolution analysis, space group analysis and merging. Additional non-isomorphism analysis is peformed and
+dataset statistics and clustering are presented in the ``xia2.multiplex.html`` report.
+For full details, see the publication at https://doi.org/10.1107/S2059798322004399 .
+
+Although xia2.multiplex will automatically determine the resolution and space group, these can be manually overridden by setting the options
+``resolution.d_min`` and ``symmetry.space_group``.
+
+One feature of xia2.multiplex is the ability to optionally trigger further processing of subsets of the data. 
+This guide provides an updated description of the different clustering options, which have been updated to use more intuitively named
+options in xia2/DIALS versions greater than v3.18, as well as incorporating some additional features developed since this initial publication.
```
jbeilstenedmands, line 17: We will need to update this to v3.22

tests/regression/test_multiplex.py
```diff
@@ -223,14 +226,61 @@ def test_proteinase_k_dose(
...
-)
-def test_proteinase_k_min_completeness(parameters, proteinase_k, run_in_tmp_path):
+def test_proteinase_k_coordinate_clusters(proteinase_k, run_in_tmp_path):
+    expts, refls = proteinase_k
+    parameters = [
+        "clustering.output_clusters=True",
+        "clustering.method=coordinate",
+        "symmetry.space_group=P422",
+    ]
+    command_line_args = parameters + expts + refls
```
jbeilstenedmands, line 236: I think this is unstable from memory, let's use expts[:-1] and refls[:-1] and then will need to reduce the assert in L243 to 7.

src/xia2/cli/cluster_analysis.py
```diff
@@ -17,64 +16,54 @@
...
+)
 from xia2.XIA2Version import Version
 
 logger = logging.getLogger("xia2.cluster_analysis")
 
+xia2_cluster_phil_scope = """\
+clustering
+  .short_caption = "Clustering"
+{
+  output_correlation_cluster_number = None
```
jbeilstenedmands, line 36: We should add a test for these options for cluster_analysis.

amyjaynethompson: Have added a test to check for this

tests/regression/test_multiplex.py
amyjaynethompson: Have put this in - lets hope it fixes the tests :)

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -337,9 +326,12 @@ def __init__(self, experiments, reflections, params):
         if all([params.symmetry.laue_group, params.symmetry.space_group]):
             raise ValueError("Can not specify both laue_group and space_group")
 
+        # self._params.symmetry.cosym.weights = "standard_error"
```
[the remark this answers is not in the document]
amyjaynethompson: Comments removed

jbeilstenedmands (approved):

PR #786, opened 2024-04-05
Cosym tolerance tailoring

amyjaynethompson:
Some auto-processed multi-crystal VMXi data from a user which was accessed through ispyb (processed with dials3.17) displayed very different multiplex clustering compared to current dials. It showed two distinct clusters in the correlation clustering on dials3.17, whereas current dials only produced one main cluster. After some investigating, it was found that the tightened tolerances in cosym rejected the data sets from this second cluster as being too different in the unit cell. We were able to manually process these clusters individually, and our user reported that after refinement the two clusters showed interesting structural differences. Therefore, we envision some use cases in multiplex where we may want to loosen these tolerances for more automated processing of clusters. These parameters `relative_length_tolerance` and `absolute_angle_tolerance` have been made accessible through multiplex in this PR without changing the default behaviour.

### DIFF
--- a/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
+++ b/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
@@ -127,7 +127,10 @@ symmetry
     .short_caption = "Resolve indexing ambiguity"
   cosym {
     include scope dials.algorithms.symmetry.cosym.phil_scope
-    include scope dials.command_line.cosym.phil_scope
+    relative_length_tolerance = 0.05
+      .type = float(value_min=0)
+    absolute_angle_tolerance = 2
+      .type = float(value_min=0)
   }
   laue_group = None
     .type = space_group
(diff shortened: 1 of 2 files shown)


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -127,6 +127,10 @@
     .short_caption = "Resolve indexing ambiguity"
   cosym {
     include scope dials.algorithms.symmetry.cosym.phil_scope
+    relative_length_tolerance = 0.05
```
jbeilstenedmands, line 130: I think it would be useful to add a help string for the two new parameters, something like "a fractional tolerance used to determine whether cells are similar during the cosym symmetry analysis"?

amyjaynethompson: I realised Graeme had also made the same comment on dials.correlation_matrix :laughing: I have now made the .help the same across both scripts.

jbeilstenedmands (approved):
Thanks for adding this feature!

Issue #787, opened 2024-04-16
Recalculate batch offsets after cosym


biochem-fan:
@graeme-winter 

Now that [dials.cosym filters out datasets that can't be mapped through a consistent best cell, rather than exiting](https://github.com/dials/dials/commit/eff2f55415c2c17addd878587d2c28bff0c71389), we have to recalculate batch offsets held by `data_manager` after running `cosym`. We cannot simply swap experiment lists.

That is, we have to call `self._data_manager._set_batches()` after this:
https://github.com/xia2/xia2/blob/28e0bc80dc7c0181c67d26e61c99fac88c8cde5b/src/xia2/Modules/MultiCrystal/ScaleAndMerge.py#L901-L910

Otherwise, we mess up batch IDs. If we are (un)lucky, the problem is manifested by
```
File "/prog/dials/modules/dials/src/dials/util/batch_handling.py", line 35, in _reduce
        assert len(set(reduced_batches)) == len(reduced_batches)
```
(i.e. duplicated batch numbers) but sometimes this leads to silent data corruption.

amyjaynethompson:
Thank-you for bringing this to our attention - this has been addressed in PR#789 and merged in. Let us know if you still have any lingering issues.

PR #789, opened 2024-04-19
Multiplex batch fix

amyjaynethompson:
To address issue #787 : 

 - As suggested, this will reset the batches in the data manger after new cosym filtering 
 - Also reset the batches after other filtering stages in multiplex (unit cell filtering and deltacchalf filtering) 

The additional batch resets on top of those suggested in #787 was in part because I had come across the same error recently, but for my datasets this seemed to be caused by the deltacchalf filtering step. After discussions with @jbeilstenedmands we thought it best to do this reset after all filtering steps to cover our bases.


jbeilstenedmands (approved):
Thanks for adding this fix.

I tested this change on an example dataset and although I wasn't able to force a crash, the batch numbering/labeling after filtering in the `xia2.multiplex.html` is still correct so looks good to me.

PR #793, opened 2024-05-31
Merge in corr mat

amyjaynethompson:
Have xia2.multiplex call from dials.correlation_matrix - all relevant clustering code has been migrated over.

Have also updated xia2.cluster_analysis to call from dials.correlation_matrix.

Also fixes a bug found in xia2.multiplex where the tables in the html listing all clusters were found to have incorrect multiplicity and completeness statistics (indices were shifted by 1, so while the listed cluster ids were correct, the statistics were calculated based on different dataset ids).


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -1024,9 +1042,12 @@ def report(self):
         )
 
     def cluster_analysis(self):
-        mca = self._mca.cluster_analysis()
-        self._cos_angle_clusters = mca.cos_angle_clusters
-        self._cc_clusters = mca.cc_clusters
+        # mca = self._mca.cluster_analysis()
```
jbeilstenedmands, line 1045: Just a bit of tidying required here to remove the commented out lines

src/xia2/cli/cluster_analysis.py
```diff
@@ -143,68 +158,78 @@ def run(args=sys.argv[1:]):
         reflections_all.extend(refl)
     reflections_all.assert_experiment_identifiers_are_consistent(experiments)
 
+    data_manager = DataManager(experiments, reflections_all)
```
jbeilstenedmands, line 161: Again, it looks like here we just need the data manager for copying and selecting, I think it would be better to bypass the need for a data manager and just work on the list of single reflection tables by rewriting the output_cluster function to work on reflections, then we would also not need to create `reflections_all`.

jbeilstenedmands (changes requested):
Thanks for migrating the code, looks good overall, just a few places where I think the memory usage can be managed better.

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
amyjaynethompson: Done :)

src/xia2/Modules/MultiCrystalAnalysis.py
```diff
@@ -107,41 +109,50 @@ def unit_cell_clustering(
...
+        filtered_ids_to_identifiers_map = copy.deepcopy(
+            self._data_manager.ids_to_identifiers_map
         )
 
-        self._cc_cluster_json = mca.to_plotly_json(
-            mca.cc_matrix, mca.cc_linkage_matrix, labels=labels
-        )
-        self._cc_cluster_table = mca.as_table(mca.cc_clusters)
+        for i in self._data_manager.experiments:
+            dup = copy.deepcopy(self._data_manager)
```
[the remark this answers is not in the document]
amyjaynethompson: Have changed this to avoid duplicating the data manager

src/xia2/cli/cluster_analysis.py
amyjaynethompson: Updated the output_cluster function to write out based on experiments and reflections to avoid needing a data manager and reflections_all

src/xia2/Modules/MultiCrystal/__init__.py
```diff
@@ -1,478 +0,0 @@
...
-        (
-            correlation_matrix,
-            linkage_matrix,
-        ) = self.compute_correlation_coefficient_matrix()
-
-        cos_angle_matrix, ca_linkage_matrix = self.compute_cos_angle_matrix()
-
-        d = self.to_plotly_json(correlation_matrix, linkage_matrix, labels=labels)
-
-        with open("%sintensity_clusters.json" % self._prefix, "w") as f:
```
jbeilstenedmands, line 106: We are now missing this in the output files, so need to add back into `cluster_analysis.py` somewhere? Also was this output by multiplex too?

amyjaynethompson: Have added output of cluster.json files for both `cluster_analysis.py` and `multiplex` - found a bug though when running the multiplex tests - have fixed and made a PR in dials to address this. Will need to be merged so these tests can be rerun.

jbeilstenedmands: I had misunderstood that this was not quite the same as what we are outputting now - the data in these json files were the plotly data, which is also contained in `xia2.multiplex.json`. But it's good to have the new output json which can be useful for follow on analysis 🙂

jbeilstenedmands (approved):
Thanks for addressing the comments.

This does slightly change the output files from multiplex - specifically there is no longer `intensity_clusters.json` and `cos_angle_clusters.json` (the data of which are contained in `xia2.multiplex.json`). And then the `xia2.multiplex_clusters.json` provides new information. It would be worth capturing this in the merge commit message.

p.s. you can ignore the "ignore merge conflicts" failing test - that always happens for merge requests from a forked xia2 repo.

PR #795, opened 2024-06-06
Replace isort/flake8/black with ruff

ndevenish:


.azure-pipelines/azure-pipelines.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
       # Run flake8 validation on a shallow clone
```
benjaminhwilliams, line 65:
```suggestion
      # Run Ruff validation on a shallow clone
```

ndevenish: Ah, dammit, I text-searched but forgot that ripgrep ignores hidden files/folders by default, separately from `--no-ignore`.

PR #814, opened 2024-09-27
Dimensionallity optimisation

amyjaynethompson:
Changes to be compatible with dials PR #2743. 

Dimension optimisation for cos-angle clustering is now able to be done within xia2.multiplex and xia2.cluster_analysis.


src/xia2/cli/cluster_analysis.py
```diff
@@ -80,8 +72,6 @@
 output {
   log = xia2.cluster_analysis.log
     .type = str
-  json = xia2.cluster_analysis.json
```
jbeilstenedmands, line 83: This shouldn't be removed or the program crashes?

amyjaynethompson: This line is back in, and all scripts now calls from dials.algorithms.correlation.analysis.working_scope rather than dials.command_line.correlation.phil_scope

ndevenish:
Above commit to try and get new CI to run when it unexpectedly didn't.

PR #823, opened 2024-11-12
Cluster name change

amyjaynethompson:
Edits to xia2.multiplex to address issues running in autoprocessing pipelines. 
 - make important cluster output files have unique names
 - prevent scaling and merging coordinate clusters if they contain all datasets


jbeilstenedmands (approved):
Thanks, this gives the files a unique name but maintains the existing directory structure 👍

Issue #818, opened 2024-10-18
 xia2/tests /regression/test_ssx.py on Windows


bwithd:
it uses * directly in the file path which is not valid on Windows.
Error: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\tqt73344\\.cache\\dials_data\\cunir_serial\\merlin0047_1700*.cbf

dagewa:
Should be fixed by #824, although it's not actually the same error that is fixed (I did not see the problem with `*`, but I did find other issues). Please re-open if problems are seen again

PR #777, opened 2024-02-08
Fix Windows tests

dagewa:
WIP branch to resolve issues with tests on Windows. First commit ensures that `".bat"` is added to the dispatcher names when called with `subprocess.run` with `shell=False`.

PR #824, opened 2024-11-20
Resolve command paths and set UTF-8 mode for tests on Windows

dagewa:
#777 assumed all DIALS command paths end with the extension `.bat` on Windows.

Not so, in some circumstances (cmake build?) the commands are actually `.exe`. Procrunner always did this properly, so let's borrow the `windows_resolve` function from there.

First step in trying to deal with #818


dagewa:
@Anthchirp, this copies your function from procrunner, with a note in the docstring. Is this attribution sufficient?

ndevenish:
Does `shutil.which` work? We've started using it elsewhere and it looks like it should "do the right thing" in [the documentation](https://docs.python.org/3/library/shutil.html#shutil.which).

graeme-winter:
> Does `shutil.which` work? We've started using it elsewhere and it looks like it should "do the right thing" in [the documentation](https://docs.python.org/3/library/shutil.html#shutil.which).

I note we use it in the tests e.g. 

```
def test_from_image_files_implicit_chunks(dials_data, tmp_path):
    image_files = sorted(
        dials_data("centroid_test_data", pathlib=True).glob("centroid*.cbf")
    )

    # Import from the image files
    result = subprocess.run(
        [
            shutil.which("dials.import"),
            "output.experiments=imported.expt",
            "split=3",
        ]
        + image_files,
        cwd=tmp_path,
        capture_output=True,
    )
    assert not result.returncode
    assert (tmp_path / "imported.expt").is_file()
    exp = load.experiment_list(tmp_path / "imported.expt")
    assert len(exp) == 3
```

ndevenish:
I'm left more confused; it looks like `windows_resolve` is using `shutil.which`... with special handling if you have passed in the full path already. Wouldn't you _not_ need to find the executable in that case?

dagewa:
Good point, I barely looked at `windows_resolve`, just copied it in because I knew this worked in procrunner. Happy to change this to a bare `shutil.which`, though I suppose we really need to do this _everywhere_

dagewa:
Ok, the special handling is so you can pass `C:\\some\\path\\dials.thing` and it would resolve to the `.exe`, `.bat` or whatever, whereas `shutil.which` on its own won't, apparently. We probably don't need that special handling in usage across DIALS, xia2 and so on

dagewa:
Ok, switched to `shutil.which`, but pre-commits showed a mypy type issue: `shutil.which` might return `None`. So far I only fixed that in the three places it complained about, where functions had type annotations.

dagewa:
All the other places are tests, so I think this is fine

src/xia2/Driver/DriverHelper.py
```diff
@@ -1,11 +1,14 @@
...
 
+import logging
 import os
 import random
 import signal
 import stat
 import string
 
+logger = logging.getLogger("xia2.Driver.DriverHelper")
+
```
ndevenish: This is leftover from old implementation?
```suggestion
import os
import random
import signal
import stat
import string

```

ndevenish:
I think I'd be tempted to reduce the `shutil.which` failure handlings to plain `assert`. It's a pity there is no builtin for some equivalent of `shutil.which("xia2") or raise(RuntimeError("No xia2"))`

src/xia2/Driver/DriverHelper.py
dagewa: Yes, it was residual from `windows_resolve`

dagewa:
```
assert (cmd := shutil.which("dials.import"))
```
?

ndevenish:
FWIW assigning inline in the assert will break if python is ever run with PYTHONOPTIMIZE / `-O`

```
% python3 -O
Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:54:21) [Clang 16.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> assert (a := 3)
>>> a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'a' is not defined
>>>
```

PR #829, opened 2024-11-26
`STRONG_PIXEL` --> `SIGNAL_PIXEL`

dagewa:
As per [changes](https://xds.mr.mpg.de/html_doc/Release_Notes.html) in XDS BUILT=20240712


ndevenish:
Does this break with the XDS-revert-shenanigans, or was this already present in the old version we rolled back to?

ndevenish:
People with commercial licences can apparently run old XDS. Is that something that we want to worry about?

PR #841, opened 2025-04-01
Add pca plots

amyjaynethompson:
Pull through the pca plots from dials PR #2885 into xia2.multiplex output


jbeilstenedmands (approved):
Looks good, once CI tests pass after corresponding dials PR is merged (they pass locally for me 👍 ).

PR #845, opened 2025-05-12
Added new features to xia2.multiplex for exporting unmerged reflectio…

### LINKED ISSUE #643: xia2.multiplex: export unmerged reflections in mmcif format
This would be useful if we are to encourage users to deposit unmerged reflection files.

C.f. https://github.com/xia2/xia2/pull/537

TiankunZhou:
…ns in mmcif format, fixes #643 (#643)


src/xia2/Modules/MultiCrystal/data_manager.py
```diff
@@ -205,6 +205,18 @@ def export_unmerged_wave_mtz(self, wl, prefix, d_min, wavelength_tolerance):
             return params.mtz.hklout
         return None
 
+    def export_unmerged_wave_mmcif(self, wl, prefix):
+        data = self.data_split_by_wl[wl]
+        nn = len(self.wavelengths)
+        fmt = "%%0%dd" % (math.log10(nn) + 1)
+        index = sorted(self.wavelengths.keys()).index(wl)
+        params = export.phil_scope.extract()
```
jbeilstenedmands, line 213: I think here we should also set the d_min value to make sure we are exporting the same resolution range. The dials export_mmcif code actually uses the `mtz.d_min` parameter to filter the data (yes that is unexpected and a bad design).

src/xia2/Modules/MultiCrystal/data_manager.py
```diff
@@ -246,6 +258,13 @@ def export_unmerged_mtz(self, filename, d_min=None, wavelength_tolerance=0.0001)
         params.intensity = ["scale"]
         export.export_mtz(params, expt_to_export, [self._reflections])
 
+    def export_unmerged_mmcif(self, filename):
+        params = export.phil_scope.extract()
+        expt_to_export = copy.deepcopy(self._experiments)
+        params.mmcif.hklout = filename
```
jbeilstenedmands, line 264: Let's set d_min here too.

jbeilstenedmands (changes requested):
Thanks for implementing this feature, I think it is nearly there, see the comment on setting d_min.
I think there are a few places in `test_multiplex.py` where clustering is done and the existence of output files are checked e.g. L217 and other places, where we also need to check for the mmcif file existence.
In testing, I noticed that dials does not export exactly the same reflections in mtz and mmcif, but I think that should be fixed separately in dials, see https://github.com/dials/dials/issues/2921

TiankunZhou:
Changes implements, please check when it is convenient for you.

PR #852, opened 2025-05-27
Fix processing of Nexus Singla datasets with the XDS pipeline

mpks:
Allow xia2 to read Nexus Singla files and process them with the XDS pipeline.

The XDS processing requires two files (a master and a data file). These two files are needed for the Durin plugin. The metadata for the Singla datasets is in the Nexus file. xia2 will read the Nexus file and get the metadata, but XDS still needs the master and the data file.


src/xia2/Wrappers/XDS/XDS.py
```diff
@@ -472,7 +474,13 @@ def template_to_xds(template):
...
         # that the word master is in there, so... otherwise can get complicated
         # side-effects when people have a folder named e.g. data_200.
-        return template.replace("master.h5", "??????.h5")
+
+        if template.endswith("master.h5"):
+            return template.replace("master.h5", "??????.h5")
+        elif template.endswith(".nxs"):
+            return template.replace(".nxs", "_??????.h5")
+        else:
+            return template
```
ndevenish, line 483: Obviously `re` needs to be added also
```suggestion
      return re.sub(r"_master\.h5$|\.nxs$", "_??????.h5", template)
```

but - do we know why we apparently need to do this renaming? durin limitation?

newsfragments/852.bugfix
```diff
@@ -0,0 +1,4 @@
+Fix processing of Nexus Singla datasets with the XDS pipeline
+
+Fix passing of the xia2 Phil parameters to the XDS wrapper
```
ndevenish, line 3: I'm surprised that this didn't cause more problems. What specifically did this break?

src/xia2/Wrappers/XDS/XDS.py
mpks: Yes, this looks better. I missed that ".nxs" can also appear in the template name, not just at the end. 

The given template name is copied into the XDS.INP file. I haven't looked into the Durin code, but I guess that is the reason. This part of the xia2  is written only for Eiger detectors, which produce a data file and a master file. The Durin plugin is probably written to deduce image indexing from the `_??????` part. I tried running the XDS.INP script with just the Nexus file (.nxs), but it fails. Fixing this on the Durin side would require more work, so this looked easier.

newsfragments/852.bugfix
mpks: The Phil parameters from the command line were not passed to the XDS wrapper. For example, you can specify XDS options for integration with `xds.integrate`. All these options would just be ignored, and XDS would run with some default ones. If your data got processed correctly, you wouldn't get any error, so you wouldn't notice.

src/xia2/Wrappers/XDS/XDS.py
```diff
@@ -458,7 +460,8 @@ def template_to_xds(template):
     if is_hdf5_name(template):
         # Given (e.g.) XYZ_master.h5 and data files XYZ_data_00000[0-9].h5
         # XDS expects the template XYZ_??????.h5
-        assert template.endswith("master.h5"), template
+
+        # assert template.endswith("master.h5"), template
```
ndevenish, line 464: Please just remove this rather than leaving it in, commented

dagewa:
I tested using https://proteindiffraction.org/project/8pqk/ with
```
xia2 pipeline=3dii image=/data/8pqk/data/08AT12_ds1_1_master.h5
```
this succeeds on `main` but unfortunately fails on this branch:
```
Environment configuration...
Python => /home/fcx32934/sw/cctbx/conda_base/bin/python
CCTBX => /home/fcx32934/sw/cctbx/conda_base/lib/python3.13
CCP4 => /home/fcx32934/sw/ccp4-9
CCP4_SCR => /tmp/tmplntkodrc
Starting directory: /tmp/8pqk-branch
Working directory: /tmp/8pqk-branch
Free space:        145.07 GB
Host: nedvedice
XIA2 3.9.dev201+g303f959f
DIALS 3.dev.1338-gc46083d8f
CCP4 9.0.009
Command line: xia2 pipeline=3dii image=/data/8pqk/data/08AT12_ds1_1_master.h5
Project directory: /tmp/8pqk-branch
-------------------- Spotfinding SWEEP1 --------------------
Error: /tmp/8pqk-branch/DEFAULT/NATIVE/SWEEP1/index/4_spot_xds.refl
xia2.support@gmail.com
```

dagewa:
Yeah, looks like it's processing fine now :+1:

jbeilstenedmands:
Ok, I imagine it was the changes to `dials.import_xds`, which required changes in the wrappers in xia2.

PR #858, opened 2025-06-04
add debug log and report timings

amyjaynethompson:
Debug log added to xia2.multiplex including timings of various steps 

Future PRs will include wrapping more of the steps to be output in the timings log but this at least sets the foundation for future changes


jbeilstenedmands (approved):
Thanks for adding this, it will be very useful 👍

PR #860, opened 2025-06-04
Use the dials delta-cchalf code for the xia2.multiplex report.

jbeilstenedmands:
In profiling the runtime of xia2.multiplex on a reasonably sized dataset (240 sweeps x 20 degrees, 4M reflections), a large fraction of time was spent in generating the multiplex report (400s / 1070s), almost all of which was in the delta cc half analysis (380s). The dials implementation seems much more efficient than the separate xia2 implementation, and by updating the multiplex report to call the dials code, this part of the analysis took only 15s.

Depends on https://github.com/dials/dials/pull/2947/


jbeilstenedmands:
I should have declared that in my above timings I ran xia2.multiplex with the `symmetry.space_group` option, so the cosym analysis (60s) was quicker than it would have been if doing the fully cosym analysis.

jbeilstenedmands:
Note that this PR now introduces a new style of wrapper, to wrap a dials progam 'functional interface' (rather than command line interface). Awaiting further discussion on this before merging in this PR, as we will want to use this approach for other unwrapped bits of `xia2.multiplex` and likely also the `xia2.ssx` code.

amyjaynethompson (commented):
Looks great! Nice that the dials version is being printed to the logs, and tested the diff_phil function with a case that did have differences - seemed to work as expected :) 

I did notice that the stream from `x_dials.compute_delta_cchalf.log` was also being output to the main multiplex log - I think it just needs `dials_logger = logging.getLogger("dials")` and `dials_logger.handlers.clear()` in `multiplex.py`?

PR #863, opened 2025-06-12
Merge functional wrapper

amyjaynethompson:
Following on from JBE's addition of functional wrappers to xia2, here is one to wrap the dials.merge functionality within multiplex (for both single and multi-wavelength datasets). 

This was prioritised to further tidy up xia2.multiplex.log. Now, only a summary of the merging process is retained here, and most of the merge log stream is output to separate files following current xia2.multiplex-style.

Note, as it currently stands, existing logic retained (ie setting assess_space_group to False). 

Also added in a new test for the wrapper, and OPTICS clustering results are also now manually output.


src/xia2/Wrappers/Dials/Functional/Merge.py
```diff
@@ -0,0 +1,59 @@
...
+
+class Merge:
+    def __init__(self, working_directory: Path | None = None) -> None:
+        if working_directory:
+            self._working_directory = working_directory
+        else:
+            self._working_directory = Path.cwd()
+
+        self.params: libtbx.phil.scope_extract = phil_scope.extract()
+        self.working_directory = working_directory
```
jbeilstenedmands, line 28: Looks like a leftover line of code that is not longer needed?

src/xia2/Wrappers/Dials/Functional/Merge.py
```diff
@@ -0,0 +1,59 @@
...
+    def set_wavelength_tolerance(self, wavelength_tolerance: float) -> None:
+        self.params.wavelength_tolerance = wavelength_tolerance
+
+    def set_r_free_params(self, r_free_params: libtbx.phil.scope_extract) -> None:
+        self.params.r_free_flags = r_free_params
+
+    def set_assess_space_group(self, assess_space_group: bool) -> None:
+        self.params.assess_space_group = assess_space_group
+
+    def set_output_filename(self, filename: str) -> None:
```
jbeilstenedmands, line 43: My idea for the new-style wrappers would be to use python properties rather than get and set methods. This is not applicable to setting properties on params (I think), but for other class members like output_filename we would have it as a private member (`self._output_filename`) and define it as a `@property` and define a method with a setter decorator.

jbeilstenedmands: Although that could look a bit inconsistent if we have lots of "set_" methods for things that are set on params and then properties for the rest 🤔 .

jbeilstenedmands (changes requested):
Thanks for adding this, the multiplex log looks much neater and the separate dials.merge log is a useful addition 👍 

We should keep the `export_merged_wave_mtz` and `export_merged_mtz` logic within standalone functions, but not necessarily tied to the `data_manager`, and add more thorough testing of the Merge wrapper.
But functionally it is good, so once the comments are addressed this will be good to go.

src/xia2/Wrappers/Dials/Functional/Merge.py
amyjaynethompson: Yep! Has been removed

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -715,14 +747,24 @@ def scale_cluster(
...
+                    index = sorted(data_manager.wavelengths.keys()).index(wl)
+                    name = f"{output_name}_WAVE{fmt % (index + 1)}.mtz"
+
+                    wave_merge.set_output_filename(name)
+                    wave_merge.set_d_min(scaled.d_min)
+                    wave_merge.set_wavelength_tolerance(
+                        self._params.wavelength_tolerance
+                    )
+                    wave_merge.set_r_free_params(self._params.r_free_flags)
+                    wave_merge.set_assess_space_group(False)
```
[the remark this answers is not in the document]
amyjaynethompson: Have made it default in the wrapper

tests/Wrappers/Dials/Functional/test_merge.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_non_default_parameters(lcy_data, run_in_tmp_path):
+    expts, refls = lcy_data
+    merge = Merge()
+    merge.set_d_min(2.4)
+    merge.set_wavelength_tolerance(0.002)
+    merge.set_r_free_params(params.r_free_flags)
+    merge.use_xpid = False
+    merge.assess_space_group = True
```
jbeilstenedmands, line 42:
```suggestion
    merge.set_assess_space_group(True)
```

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -1173,6 +1178,38 @@ def cluster_analysis(self) -> None:
         self._cos_angle_clusters = self._mca.cos_clusters
         self._cc_clusters = self._mca.cc_clusters
 
+    def export_merged_mtz(
+        self,
+        expts: ExperimentList,
+        refls: flex.reflection_table,
+        file_name: str,
+        d_min: str,
```
jbeilstenedmands, line 1186:
```suggestion
        d_min: float,
```

amyjaynethompson: Fixed to float

tests/Wrappers/Dials/Functional/test_merge.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_non_default_parameters(lcy_data, run_in_tmp_path):
+    expts, refls = lcy_data
+    merge = Merge()
+    merge.set_d_min(2.4)
+    merge.set_wavelength_tolerance(0.002)
+    merge.set_r_free_params(params.r_free_flags)
+    merge.use_xpid = False
+    merge.set_assess_space_group = True
```
jbeilstenedmands, line 42:
```suggestion
    merge.set_assess_space_group(True)
```

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -1173,6 +1178,38 @@ def cluster_analysis(self) -> None:
...
+        self,
+        data_manager: DataManager,
+        wl: float,
+        output_name: str,
+        d_min: float,
+    ) -> str | None:
+        data = data_manager.data_split_by_wl[wl]
+        if data["expt"]:
+            fmt = "%%0%dd" % (math.log10(len(self._data_manager.wavelengths)) + 1)
+            index = sorted(data_manager.wavelengths.keys()).index(wl)
```
ndevenish, line 1205:
```suggestion
        wavelength: float,
        output_name: str,
        d_min: float,
    ) -> str | None:
        data = data_manager.data_split_by_wl[wavelength]
        if data["expt"]:
            fmt = "%%0%dd" % (math.log10(len(self._data_manager.wavelengths)) + 1)
            index = sorted(data_manager.wavelengths.keys()).index(wavelength)
```
Unless using `wl` is an _extremely_ common xia2 idiom?

src/xia2/Wrappers/Dials/Functional/Merge.py
ndevenish: I agree with properties in general but I suspect that these getters/setters are a very heavily embedded idiom...

Issue #847, opened 2025-05-14
xia2.multiplex logging standardisation


jbeilstenedmands:
The `xia2.multiplex` log file/stdout is rather a different style to other xia2 programs (`xia2` and `xia2.ssx`), which in my opinion makes it more difficult to follow and interpret. Some of the programs run don't really feature in the log at the moment, or don't have dedicated log files (e.g. `dials.correlation_matrix`).

As part of standardising, I would expect banners of the program names, summary output of the key results from each algorithm step and clear, concise merging stats tables. Also, adding a debug log file including a timing report would be very useful for targeting effort for performance improvements in the multiplex code, as we currently don't have a clear idea of the time spent in the multiplex wrapper code relative to the underlying dials programs.

Changes to be discussed and presented to users for feedback before merging a change.

jbeilstenedmands:
A user has noted that the unit cell value and estimated sds are not reported in the log, I think this would be a useful thing to add into the main log after each call of dials.two_theta_refine (main process and for any cluster processing).

PR #877, opened 2025-08-05
Logging additions

amyjaynethompson:
Following user comment from #847, refined unit cell from two-theta refine now reported in xia2.multiplex.log (for overall dataset as well as individual clusters). Also added in a line noting what symmetry was used if set by the user (without this, the ----- Applying consistent symmetry ----- banner in the main log would report nothing before moving to ----- Scaling -----).


jbeilstenedmands (approved):
Thanks for adding this, looks good.

If you merge the main branch into this branch, you should find the tests will pass.

PR #879, opened 2025-08-07
Parallelise scaling and merging of output clusters

amyjaynethompson:
One easy way to improve the speed of xia2.multiplex is to scale and merge any identified clusters in parallel, rather than sequentially. This does work on my data (and greatly speeds things up!), but more testing/tidying is needed.... However, I am about to go on holiday so leaving this here as a draft :)


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -587,6 +577,7 @@ def run(self) -> None:
                 self._mca.significant_coordinate_clusters
             )
             count = 0
+            subclusters = []
```
jbeilstenedmands, line 580: It looks like we're initializing the subclusters list twice, do we need to define it once before all the different cluster code branching?

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -602,11 +593,39 @@ def run(self) -> None:
...
-                    self._data_manager, cluster_dir, cluster_identifiers, s_c
+                subclusters.append(
+                    (cluster_dir, "coordinate", cluster_identifiers, s_c)
                 )
                 count += 1
 
+        with (
+            record_step("dials.scale(parallel)"),
+            concurrent.futures.ProcessPoolExecutor(
+                max_workers=self._params.nproc
```
jbeilstenedmands, line 604: I think we want to limit max workers to be `min(self._params.nproc, len(subclusters))` to avoid spinning up more ProcessPool's than needed? Also, do we then need to restrict nproc to 1 for the programs used within `scale_and_report_cluster`, to avoid further process spawning?

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -517,95 +520,56 @@ def run(self) -> None:
...
-                ]
-                self._scale_and_report_cluster(
-                    self._data_manager, cluster_dir, cluster_identifiers, s_c
-                )
-                count += 1
+            cluster_futures = {
+                pool.submit(
+                    self._scale_and_report_cluster,
+                    self._data_manager,
+                    cluster[0],
```
jbeilstenedmands, line 551: Indexing into a lust for the subcluster attributes looks unclear and error prone. I think we need to define a simple SubCluster dataclass with the relevant named attributes. Then `get_subclusters` will have a much simpler type signature and everything will be a bit clearer in several functions in `cluster_analysis.py`

jbeilstenedmands:
btw the code structure looks much better with more cleanly separated functions in cluster_analysis now 👍

jbeilstenedmands (approved):
Thanks for addressing the comments and for implementing this improvement, this looks good to go in now.

PR #881, opened 2025-09-05
Output extra merged file in xia2.ssx at the suggested resolution limit.

jbeilstenedmands:
xia2.ssx: When the suggested resolution limit is lower than the resolution range, output two merged files, merged.mtz and merged_full.mtz.

In cases where the resolution extends beyond the limit of the data, just output the merged.mtz file.

i.e. a user would be taking the merged.mtz file by default. If it has been cut and you want the full resolution range, then the merged_full.mtz file is also available.


amyjaynethompson (approved):
Extra logging messages are helpful and clear, and I think this will be very useful for users :)

PR #884, opened 2025-09-23
revert to serial cluster processing until multiplex memory usage improved

amyjaynethompson:
Recent addition of parallel scaling/merging for individual clusters broke for large datasets (OOM faults...). Here we attempt to restructure to avoid memory copies of large data classes (data_manager, MultiCrystalScale) by making mini-data managers before the pool and moving more methods to static.

### DIFF
--- a/src/xia2/Modules/MultiCrystal/data_manager.py
+++ b/src/xia2/Modules/MultiCrystal/data_manager.py
@@ -26,7 +26,7 @@ class DataManager:
         self,
         experiments: ExperimentList,
         reflections: flex.reflection_table,
-        batch_offset_list: list[int] = [],
+        batch_offset_list: list[int] | None = None,
     ):
         self._experiments = experiments
         self._reflections = reflections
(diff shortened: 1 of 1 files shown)


src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -552,12 +562,14 @@ def run(self) -> None:
                 cluster_futures = {
                     pool.submit(
                         self._scale_and_report_cluster,
-                        self._data_manager,
+                        self._params,
+                        item.data_manager,
```
jbeilstenedmands, line 566: Could we just pass the cluster objects as arguments to scale_and_report_cluster, rather than a long list of arguments that are the cluster properties?

src/xia2/Modules/MultiCrystal/data_manager.py
```diff
@@ -22,7 +22,12 @@
 
 
 class DataManager:
-    def __init__(self, experiments: ExperimentList, reflections: flex.reflection_table):
+    def __init__(
+        self,
+        experiments: ExperimentList,
+        reflections: flex.reflection_table,
+        batch_offset_list: list[int] = [],
```
jbeilstenedmands, line 29: It is dangerous to use a mutable default argument in a function in python, instead it should be default initialised to None like `batch_offset_list: list[int] | None = None`

PR #887, opened 2025-10-15
xia2.multiplex - add shelx export for small molecule

amyjaynethompson:
xia2.multiplex is also very useful for small molecule data (used on I19 and VMXm at Diamond regularly). Since small molecule refinement tends to happen in a .hkl/.ins format rather than .mtz format, this PR adds the dials.export functionality to multiplex to export SHELX style files. This is currently triggered via a new section of Phil parameters - small_molecule.shelx_output = True. In the future this could be changed to just small_molecule = True when other improvements (such as symmetry handling) are complete.

### DIFF
--- a/newsfragments/887.feature
+++ b/newsfragments/887.feature
@@ -1,1 +1,1 @@
-``xia2.multiplex``: Included option to export SHELX files for small molecule crystallography
+``xia2.multiplex``: Added option (small_molecule.composition=) to export SHELX files for small molecule crystallography


dagewa:
Hi Amy, this is great :+1: I could not immediately see from the code whether it is using the DIALS functionality in `dials/util/export_shelx.py` or other SHELX export code like that in `xia2.to_shelx`. I seem to remember that the DIALS version is better, perhaps because it maximises precision in f8.2 format.

newsfragments/887.feature
```diff
@@ -0,0 +1 @@
+``xia2.multiplex``: Included option to export SHELX files for small molecule crystallography
```
jbeilstenedmands:
```suggestion
``xia2.multiplex``: Added option (small_molecule.composition=) to export SHELX files for small molecule crystallography
```

PR #894, opened 2025-11-07
make clustering output of xia2.multiplex consistent with dials.correlation_matrix

amyjaynethompson:
I found an example where the OPTICS clusters coming out of xia2.multiplex (run on scaled data) and dials.correlation_matrix on the scaled.{expt,refl} files were significantly different (different numbers of reflections were also being read). These two situations should behave the same, especially for consistency of the user 

**xia2.multiplex:**
Cluster 0 contained 10 datasets, and Cluster 1 contained 12 datasets
**dials.correlation_matrix on scaled.{expt,refl}:**
Cluster 0 contained 29 datasets, and Cluster 1 contained 12 datasets

Turns out, exporting the merged and unmerged mtz files altered the reflection table - passing in a copy of the table to these functions means that xia2.multiplex clustering on scaled data behaves the same as dials.correlation_matrix on scaled.{expt,refl} files.


jbeilstenedmands (approved):
Thanks for the fix!

PR #914, opened 2026-03-27
make cluster graphs in html clickable to merge custom subsets of data

amyjaynethompson:
In this PR, the clustering graphs in xia2.multiplex are now more interactive. Using the existing plotly features (such as trace highlighting and selection tools), custom subsets of data can now be clicked on to generate a multiplex command which can be copied directly into the command line (or script) to merge the selection. 

Dendrograms (correlation and cosine) can be clicked on, while the reachability plots, cosym coordinate plot, and PCA matrix plots can be selected with the lasso or box selection tool. Doing this brings up a pop-up window with the list of dataset IDs included in the selection, as well as a multiplex command making use of the 'identifiers' Phil parameter. This is done by also passing in the ids to identifiers dictionary from the data manager. This also relies on a feature included recently in xia2, where multiplex now outputs a working-params.phil file. This means that the settings of the original multiplex job can be used to reprocessing the subset of data. This was tweaked in this PR to also output absolute file paths of the initial expt/refl input files to the working-pail, to save the user needing to dig these out again. 

This also includes a new optionally outputted html which only contains relevant cluster graphs. These have also been split across different tabs for clarity. This is useful for allowing users to easily access clustering information in a stand-alone manner (plans to implement in auto-processing at Diamond). 

Clicking on dendrogram:

image

Pop up box:

image

Selecting on coordinate plot:

image

Selecting on reachability plot:

image


jbeilstenedmands (changes requested):
Thanks for adding this very useful feature, only a few minor comments to consider/address. Thanks!

jbeilstenedmands (approved):
Thanks for addressing the comments.

PR #915, opened 2026-04-02
Tidy multiplex file structure

amyjaynethompson:
A proposal for updating the multiplex output file structure to be more like xia2. Currently, all files are output to the same directory, with subfolders only made for output clusters. This can be confusing for users, and there are also many intermediate files produced by multiplex that 99% of users won't interact with or ever need. Therefore, this also includes a 'cleanup' mode, which will delete many of these intermediate files to save storage space. Of course, advanced users can keep these files by setting `output.cleanup=False`. 

Proposed File Structure: 
- DataFiles
  - scaled*.{mtz,sca,mmcif}
  - filtered*.{mtz,sca,mmcif}
  - \*cluster\*.{mtz,sca,mmcif}
- LogFiles
  - xia2.multiplex.log
  - xia2.multiplex.html
  - xia2.multiplex.debug.log
- Processing (xia2.multiplex will run here and files moved elsewhere at the end)
  - everything else
- citation information
- xia2-multiplex-working.phil 

General rules of thumb for removing intermediate files after use:
- remove intermediate .expt/.refl (except for `{scaled, filtered, models, observations}.{expt,refl}`)
- remove intermediate .json (except for final `xia2.multiplex.json`)
- keep images (ie `cluster_unit_cell_p1.png`)
- keep logs / html corresponding to specific steps (ie starting with xpid_)
- remove 'overall logs' (ie remove `dials.cosym.log` but keep `1_dials.cosym.log`) 

NOTE: not for immediate inclusion in a release -> at least at Diamond will need to properly assess how this changes auto processing infrastructure


src/xia2/cli/multiplex_filtering.py
```diff
@@ -225,22 +216,32 @@ def run(args=sys.argv[1:]):
...
+    refl_path = filter_params.input.directory / "Processing/observations.refl"
     logger.info(f"Using {expt_path} and {refl_path} as input data for filtering.")
 
     experiments = ExperimentList.from_file(expt_path, check_format=False)
     reflections = flex.reflection_table.from_file(refl_path)
 
     if not full_params.r_free_flags.reference:
         full_params.r_free_flags.reference = str(
-            filter_params.input.directory / "scaled.mtz"
+            filter_params.input.directory / "DataFiles/scaled.mtz"
```
jbeilstenedmands: I think we need to be careful here for windows compatibility, we should try to use pathlib everywhere for consistency which then also avoid the need for ever writing a forwardslash in a string for paths, i.e. we want to write `filter_params.input.directory / "DataFiles" / "scaled.mtz"`, which is then correctly resolved to the right path across platforms.

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,75 @@
...
+        data_file = pathlib.Path(filename).resolve()
+        if data_file not in self._data_files:
+            assert data_file.exists(), f"Required file {data_file} not found."
+            self._data_files.append(data_file)
+
+    def record_log_file(self, filename):
+        log_file = pathlib.Path(filename).resolve()
+        if log_file not in self._log_files:
+            if not os.getenv("PYTEST_CURRENT_TEST"):
+                # Tests fails when do below all the time because only creates 1 multiplex log for entire test
```
jbeilstenedmands: Remind me, what was the reason for this, and what did we think might fix it? Would it be worth fixing that now?

jbeilstenedmands: I think I remembered, we would need to run each main multiplex call in a subprocess in the tests?

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -559,6 +571,12 @@ def run(self) -> None:
                 self._individual_report_dicts[cluster_name] = individual_report
                 self._update_comparison_graphs(report, dict_report, cluster_name)
                 self._log_report_info(dict_report)
+                if self._params.output.cleanup:
```
jbeilstenedmands: Just looking at some of the patterns here, I think it would be better for the file handler to know whether cleanup is true or false from the start, i.e. we do filehandler.set_cleanup(self.params.output.cleanup) at the start of multiplex. Then we don't need lots of "if cleanup", and which files are kept depends on which method was used to record them. I would have maybe expected all files recorded with "record_temp_file" would be cleaned up, but this seems not to be the case?

src/xia2/cli/multiplex_filtering.py
amyjaynethompson: Have moved (hopefully) everything to path lib.

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
amyjaynethompson: Have removed this logic and switched to subprocess for calling multiplex

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
amyjaynethompson: Now have temp files and optional files -> temp files are always deleted, and optional files are based on `self._params.output.cleanup`. Made this more explicit in the file handler to avoid the if statements.

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,92 @@
...
+import contextlib
+import logging
+import pathlib
+
+logger = logging.getLogger("xia2.Handlers.Files")
+
+
+class _MultiplexFileHandler:
+    def __init__(self):
+        self._temporary_files = []
```
jbeilstenedmands: Could we add a comment above each list here, to be clear about the purpose of each list. Also a type hint for each list (`list[Path]`)?

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def set_cleanup(self, cleanup: bool):
+        self.delete_optional_files = cleanup
+
+    def cleanup(self, base_path):
+        base_path = pathlib.Path(base_path).resolve()
+
+        for f in self._temporary_files:
+            try:
+                pathlib.Path.unlink(f)
```
jbeilstenedmands:
```suggestion
                f.unlink()
```
is more idiomatic I think

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,92 @@
...
+                logger.debug(f"Deleted: {f}")
+            except FileNotFoundError as e:
+                logger.debug(f"Failed to delete: {f} ({e})")
+
+        if self.delete_optional_files:
+            for f in self._optional_files:
+                try:
+                    pathlib.Path.unlink(f)
```
jbeilstenedmands:
```suggestion
                    f.unlink()
```

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,92 @@
...
+        data_path = base_path / "DataFiles"
+        data_path.mkdir(exist_ok=True)
+        log_path = base_path / "LogFiles"
+        log_path.mkdir(exist_ok=True)
+
+        for f in self._log_files:
+            target = log_path / f.name
+            try:
+                f.rename(target)
+            except Exception as e:
```
jbeilstenedmands: Is there a more specific exception we could catch here (and below)?

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
```diff
@@ -0,0 +1,92 @@
...
+        self.delete_optional_files = False
+
+    def set_cleanup(self, cleanup: bool):
+        self.delete_optional_files = cleanup
+
+    def cleanup(self, base_path):
+        base_path = pathlib.Path(base_path).resolve()
+
+        for f in self._temporary_files:
+            try:
```
jbeilstenedmands: In here, we could add an extra safety check that all files that we are trying to delete are within the multiplex job:
```
if base_path in f.parents:
        f.unlink()
```

jbeilstenedmands (approved):
Thanks for addressing the initial review comments, this looks good now.

Just a few minor suggestions to clean up the file handler.

tests/regression/test_multiplex.py
```diff
@@ -293,27 +357,44 @@ def test_proteinase_k_hierarchical_clusters_distinct(proteinase_k, run_in_tmp_pa
...
-    with pytest.raises(SystemExit) as e:
-        run_multiplex([expts[0], refls[1]])
-    assert str(e.value) == "xia2.multiplex requires a minimum of two experiments"
+    result = subprocess.run(
+        [shutil.which("xia2.multiplex")] + [expts[0], refls[1]],
+        cwd=run_in_tmp_path,
+        capture_output=True,
+    )
+    assert result.returncode
+    assert result.stderr == b"xia2.multiplex requires a minimum of two experiments\n"
```
jbeilstenedmands: Looks like this line causes a test fail on windows:
```
[53](https://github.com/xia2/xia2/actions/runs/26506261982/job/78059324550?pr=915#step:9:255)
    def test_proteinase_k_single_dataset_raises_error(proteinase_k, run_in_tmp_path):
        expts, refls = proteinase_k
        result = subprocess.run(
            [shutil.which("xia2.multiplex")] + [expts[0], refls[1]],
            cwd=run_in_tmp_path,
            capture_output=True,
        )
        assert result.returncode
>       assert result.stderr == b"xia2.multiplex requires a minimum of two experiments\n"
E       AssertionError: assert b'xia2.multip...periments\r\n' == b'xia2.multip...experiments\n'
E         
E         At index 52 diff: b'\r' != b'\n'
E         
E         Full diff:
E         - (b'xia2.multiplex requires a minimum of two experiments\n')
E         + (b'xia2.multiplex requires a minimum of two experiments\r\n')
E         ?                                                         ++
```

jbeilstenedmands: Likely fixable with something like:
```
assert result.stderr.decode().splitlines() == [
    "xia2.multiplex requires a minimum of two experiments"
]
```

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
amyjaynethompson: Added descriptive comments and type hints

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
amyjaynethompson: Included specific exception

src/xia2/Modules/MultiCrystal/MplxFileHandler.py
amyjaynethompson: Added extra safety check

tests/regression/test_multiplex.py
amyjaynethompson: Oh yes, that works much nicer - thank you for the suggestion, have made this change.

jbeilstenedmands (changes requested):
Note this is not ready to merge yet as windows test failure is genuine. We're thinking of slightly changing the output file locations, so more work to come on this one.

amyjaynethompson:
New proposed structure:
- DataFiles
   - scaled*.{mtz,sca,mmcif}
   - filtered*.{mtz,sca,mmcif}
   - *cluster*.{mtz,sca,mmcif}
- LogFiles (use cluster folder structure if clusters)
   - *.log
   - *.png
   - *.html (ONLY FOR INDIVIDUAL DIALS PROCESSES)
- Processing (xia2.multiplex will run here and files moved afterwards)
   - *.json
   - *.expt
   - *.refl
- xia2.multiplex.log
- xia2.multiplex.html
- xia2.multiplex.debug.log
- citation information
- xia2-multiplex-working.phil

This means the primary logs are no longer being moved which prevents the main windows errors occurring.

src/xia2/Modules/MultiCrystal/ScaleAndMerge.py
```diff
@@ -1167,6 +1217,13 @@ def cosym(self) -> None:
         elif self._params.symmetry.laue_group:
             logger.info(f"Using input Laue group: {self._params.symmetry.laue_group}")
 
+        MultiplexFileHandler.record_temp_file("dials.cosym.log")
+        MultiplexFileHandler.record_optional_file(self._experiments_filename)
+        MultiplexFileHandler.record_optional_file(self._reflections_filename)
+        MultiplexFileHandler.record_optional_file(cosym.get_json())
+        MultiplexFileHandler.record_log_file(f"{cosym.get_xpid()}_dials.cosym.log")
+        MultiplexFileHandler.record_log_file(f"{cosym.get_xpid()}_dials.cosym.html")
```
jbeilstenedmands: I think we should be using the `get_html` and `get_json` methods from the program wrappers for robustness?

amyjaynethompson: Good point - have now done this. Note that it required a bug fix in the Cosym wrapper where get_html didn't actually return anything.

jbeilstenedmands (approved):
Thanks, I think this looks good to go now and saves a lot of disk space by removing the intermediate dials files. The processing directory can then be further deleted if users want to save more space. 👍

PR #918, opened 2026-04-22
Fix xia2.multiplex_filtering showing inconsistent results with xia2.multiplex

amyjaynethompson:
It was found that xia2.multiplex_filtering was actually not reproducing the filtering found in xia2.multiplex. This is due to the `models.expt` / `observations.refl` files being overwritten within multiplex. These are the files input to xia2.multiplex_filtering, as these are the files fundamentally used in xia2.multiplex for filtering. 

Multiplex no filtering:
- `models.expt` / `observations.refl` written out BEFORE scaling and symmetry determination 

Multiplex with filtering:
- after scaling, runs symmetry determination, then overwrites `models.expt` / `observations.refl` with current data in `data_manager` for use in filtering

Filtering on multiplex no filtering:
- uses `models.expt` / `observations.refl` from BEFORE scaling, meaning the resulting data has not undergone full symmetry determination (issue was originally identified as space group was not the same between filtering in multiplex and multiplex_filtering) 

Filtering on multiplex with filtering:
- results are consistent as `models.expt` / `observations.refl` have been overwritten and are NOT equivalent to the files used from a "standard" multiplex run with no filtering. 

To fix this, this PR distinguishes these two files by exporting `models_scaled.expt` and `observations_scaled.refl` regardless of whether filtering is run within multiplex. This allows xia2.multiplex_filtering to access the same files when run separately. These files are further exported within xia2.multiplex_filtering for consistency, but not exported twice in xia2.multiplex.


jbeilstenedmands:
Is there a reason why we can't just save a new `models.expt, observations.refl` after symmetry and scaling, but before filtering? Seems that this is the missing piece that should make everything else fall into place?

jbeilstenedmands (approved):
I think newsfragment now doesn't describe the change, but apart from that looks good.

PR #920, opened 2026-05-07
Add `--timing` option to dump timing database to JSON

dagewa:
For the A/B testing project we wanted a way to check the timing of individual commands run by xia2, to annotate overall memory profile plots to various stages of processing. It turns out that xia2 already records the required information in a timing "database", but this is only used for a pretty format of timings in the debug log file.

This PR introduces a new command line parameter, `--timing`, which dumps the database at the end of the xia2 run into the file `xia2-timings.json`, which can be used for further analysis (e.g. by the A/B testing dashboard).

This was added as a command line double-dash parameter rather than PHIL option, because the existing handling of the timing report happens outside `xia2_main()`, where we don't have access to the PHIL parameters.


jbeilstenedmands (approved):
Thanks, tested and looks good.
Only improvement would be to add this option to one of the regression tests, to verify that it will continue to work going forwards.

dagewa:
Thanks, good idea. I added it to `test_dials` rather than adding a new test because they take a while to run.

dagewa:
Windows build fails for some unknown reason, but nothing about this PR should be a problem on Windows, so I'm going to merge

PR #927, opened 2026-06-17
Defer FreeR_flag column assignment

### LINKED ISSUE #926: CCP4 `freerflag` doing nothing
It seems that xia2's free R options, like `free_total` and `free_fraction` are ignored, because CCP4's `freerflag` is not doing anything to its HKLIN file. I suspect this is because the `FreeR_flag` column has _already_ been set by `dials.merge`, which uses 

```
r_free_flags {
  generate = True
  fraction = 0.05
  max_free = 2000
```

and in that case `freerflag` appears to ignore its instruction to set, say, 5% of flags if that would exceed 2000 reflections. I definitely think that is a bug in `freerflag` (it should not silently ignore the instruction to set these values in the file), but also there is a xia2 problem. Probably we should be setting `r_free_flags.generate=False` so that `freerflag` actually does operate in the way it is intended in all pipelines.

graeme-winter:
Save for later for CCP4 freer_flag program: resolves #926


graeme-winter:
This is a trivial PR entirely defined by @dagewa 🤣 but thought I would remind myself how to make changes

dagewa (approved):
Confirmed - works exactly as intended. Thanks for the super quick fix!

PR #928, opened 2026-06-18
Consistently name dials html files output in multiplex/xia2 data reduction.

jbeilstenedmands:
Make all data reduction programs use `{xpid}_dials.{program}.html` naming convention.

Noticed for dials.symmetry while testing https://github.com/xia2/xia2/pull/915

The `{xpid}_scaling.html` name is always something that has annoyed me, fix that too for consistency.


amyjaynethompson (approved):
Very happy with this for consistency :) Thank you for fixing this

PR #930, opened 2026-06-29
Address multiplex space_group ambiguity in the parameter scope.

jbeilstenedmands:
Something that always gets me when running `xia2.multiplex space_group=X`:
```
Sorry: Ambiguous parameter definition: space_group = X
Best matches:
  symmetry.cosym.space_group
  symmetry.space_group
  resolution.space_group
```

However, in the correlation matrix work, we separated off the symmetry aspects from the main cosym phil scope, so to solve this we can now use that scope and copy the other two relevant items here (`lattice_group` is not copied, as we were never propagating that parameter from multiplex to cosym).

The main `space_group` parameter gets an expert level of 0, to promote it ahead of `resolution.space_group`.


amyjaynethompson (approved):
Thank you for this addition - it definitely removes ambiguity in using multiplex :)

Issue #96, opened 2017-01-19
Please make space_group=I2 work correctly


graeme-winter:
For the I2 setting of C2; actually make this work & output in I2

This is likely to be rather involved.

rjgildea:
See also http://jira.diamond.ac.uk/browse/I04_1-87

drmrw:
i19 users has also had this same problem when trying to use a non-standard setting, I2/m (C2/m space group 12). 

For testing purposes, the command line input was the following: 
xia2 small_molecule=true /dls/i19-2/data/2017/mt16568-1/DIA0917_020_TR333-1_Cu_Fbz_NH2pyz_N2_Vac-1/ xia2.settings.unit_cell=11.76,9.74,15.05,90,105.34,90 xia2.settings.space_group=I2/m

drmrw:
Can I re-open this issue as this is still being requested by i19 users.

graeme-winter:
Acknowledged - though as @rjgildea points out this can be complicated...

benjaminhwilliams:
https://jira.diamond.ac.uk/browse/I04_1-87

drmrw:
This is still important, but may already be fixed.

Cheers,

Mark.

PR #931, opened 2026-06-29
Allow `space_group=I2` (and other `mI` space groups)

dagewa:
This is an implementation of the specific proposal in https://github.com/xia2/xia2/issues/96#issuecomment-4057489028. I struggled to get this to work in xia2, so to keep things simple I have restricted this only to the `dials` pipeline. In addition, I have only handled the C2 / I2 case so far. This will not satisfy @drmrw yet, who wants at least the I2/m / C2/m case. I suppose this ought to be extended to every mC / mI alternative. However before I do that I wanted to get visibility on the approach and feedback on whether this an acceptable route for xia2. I'm still not completely familiar with the structure of xia2's internals and it is very possible that I have missed something.

This makes no attempt to provide a `best_monoclinic_beta` type feature like `dials.symmetry` has. Choosing I2 is done explicitly by passing `space_group=I2` or setting `dials.reindex_C2_to_I2=True`. This parameter acts as a flag to tell xia2 to convert to I2 at the end of processing. This avoids the problem that xia2 maps a single lattice character to space group number (https://github.com/xia2/xia2/issues/96#issuecomment-4054764968). This way, xia2 can do all processing up to scaling using C2, and then we convert to I2 just for scaling and export.

If the user provides a `unit_cell` as well as `space_group=I2` then that unit cell is converted to the C2 cell for processing, and converted back to the user's setting for scaling and export.

Thus, we have a "bolt-on" implementation of I2 handling, which seems to be the only way to do this without fundamentally altering xia2's symmetry code.


dagewa:
I have updated this to allow the choice of other `mI` space groups, which means it now works also in `small_molecule=True` mode

PR #932, opened 2026-06-29
Add small molecule symmetry determination to xia2.multiplex

jbeilstenedmands:
Trigger small-molecule symmetry analysis for xia2.multiplex via the `small_molecule.composition` parameter.

I think it would be best to trigger all small molecule options through a single parameter, although we could consider whether `small_molecule.composition` is the best way to do this


jbeilstenedmands:
@amyjaynethompson and I were just discussing what other changes might be wanted for small molecule processing, with the view of understanding what the parameter interface might need to be.

The two we thought about were the absorption level and anomalous parameters in scaling. These would require other changes first in multiplex to make some parameters "auto", but the idea was that based on the composition formula, the absorption level could be suitably set depending on if metals exist in the composition. And perhaps depending on the composition (+ wavelength?), anomalous could be auto set to true/false (main xia2 already does this with atom=X). Or maybe dials.scale should be more clever in these cases by observing what is happening with the data. In any case, the composition parameter still feels like a good unified way to indicate a small molecule dataset, and other small-molecule logic can be triggered based on that.

dagewa:
This is great, thanks! I did a test using a subset of the data sets at `/dls/i19-2/data/2025/nt43442-1/processing/crystallographic_data/5_deg_datasets/Ni_dppe/` (88 data sets). This seemed to do exactly what was expected
```
----------------- Determining space group ------------------
Space group determined by dials.symmetry: P 1 21/c 1
```
I understand the use of the `small_molecule.composition` parameter as a trigger, although there would then be a discrepancy with `xia2` itself, which uses a `small_molecule=True` meta parameter to set various things. As `small_molecule` is already a scope in `xia2.multiplex`, we can't easily replicate that interface. However, it is a scope with only one parameter, so an alternative would be to move `composition` to the `output` scope, where it is used to trigger shelx output, and change `small_molecule` to a bool. I'm not really sure what is best, but I think consistency with xia2 and `dials.symmetry` is a strong consideration.

dagewa:
It looks like `dials.export format=shelx` will apply the dummy composition `CH` if none is provided. So a possible interface would be:

- `small_molecule=True`: set various small-molecule related processing parameters, including refstat symmetry mode for `dials.symmetry` and trigger SHELX output with a dummy composition
- `small_molecule=True output.composition=CHNO`: as above, but set the provided composition for SHELX output

jbeilstenedmands:
Ok, yes I agree consistency is highly important. I think I am on board with your suggestion. For the other points I raised (`anomalous` and `absorption_level` in scaling), we could probably just default these to `anomalous=True` and `absorption_level=high` in the case of small molecule, unless explicitly set. So we would have a list of options (`composition`, `anomalous`, `absorption_level`,....) that get set when `small_molecule=True`. Setting the scaling options will be a separate piece of work to this PR and will require more involved testing.

amyjaynethompson (changes requested):
Happy with the discussed Phil suggestions, and very happy to have these changes in. The one thing I would add is that the documentation in this https://xia2.github.io/multiplex/chemical_crystallography_multiplex.html page should probably be updated as well (possibly with a note differentiating the different versions of dials?)

jbeilstenedmands:
Ok, I have now updated the Phil scope to @dagewa 's suggestion, and updated the documentation.

I had to update how multiplex-filtering gets the multiplex scope as the tests started to fail due to scope inconsistency after relocating the composition parameter - the definition from `src/xia2/cli/multiplex.py` was being duplicated, it now imports directly from `src/xia2/cli/multiplex.py`. I'm not sure if there was a reason for this beyond trying to not import from cli, maybe this now breaks something else @amyjaynethompson ?

amyjaynethompson (approved):
Looks good to me! Can confirm that `xia2.multiplex_filtering` still seems to work as expected - unsure why the Phil wasn't just imported, but doing so doesn't seem to have broken anything. Thank you for including this and updating the documentation.

PR #934, opened 2026-07-07
Add delta-cc-half filtering option to xia2.ssx

jbeilstenedmands:
dials.scale support cycles of scaling and delta-cc-half filtering, this PR adds support for triggering this option in the data reduction part of xia2.ssx.

This is triggered with the option `filtering.method=deltacchalf`, the default is for this not to be used as this could significantly increase the runtime for data reduction jobs. The `stdcutoff` parameter is also exposed as a key control of the harshness of filtering. Other options can be set through the `scaling.phil` mechanism.

It works slightly differently for scaling against a reference - as all crystal are scaled independently, we can only do a standalone delta-cc-half filter step with no further scaling as this would not change the scales for the remaining crystals - indeed dials.scale does not allow you to use the scaling and filtering options for scaling against a reference for this reason.

I will add a mention of this feature to the xia2.ssx html docs separately, after looking at them they need a bit of a refresh, so that will be a separate PR coming shortly after.


jbeilstenedmands:
After further consideration, the filtering phil scope has been updated to match xia2.multiplex and dials.scale.

To turn on deltacchalf filtering, the parameter usage is `filtering.method=deltacchalf`.

I also folded in the mean_i_over_sigma method into this method option, so now we have `filtering.method=deltacchalf Isigma`, and both can be chosen to combine the filtering methods. This changes the behaviour - previously I/sigma filtering was enabled by setting the `mean_i_over_sigma_threshold` parameter to not None, although I do not believe this parameter is being commonly used. Now that parameter is set to a default value (0.5) and you must use `filtering.method=Isigma` to trigger.

PR #940, opened 2026-07-27
multiplex: merge all pngs into html for clarity

amyjaynethompson:
To improve readability of multiplex and make file output more efficient, this PR effectively removes all pngs output by the software and migrates relevant information into the HTML file. This was inspired by a user not realising that half of their data had been clustered out by the preliminary unit cell clustering as the unit cell clustering plots in the HTML were only on the final datasets. Having pngs + HTML log output is confusing for users. 

Unit Cell Clustering:
The unit cell clustering tab in the multiplex HTML has been updated to include an extra plot of the results of the preliminary unit cell clustering. The data taken forward are highlighted and labels on subsequent graphs changed to make it clear that they are the final datasets. 

image

Multiplicity Plots:
These are already available in the data tabs in the HTML file, therefore there is no need to output the PNGs as well. The code for this is in the Report then plot_multiplicity cli program. A new optional parameter has been added to make the graph output optional. By default, the graphs will be output. 

Two Theta Refinement:
After conversations with @dagewa we have decided this probably isn't that useful for the cell refinement in xia2.multiplex, so the output of this graph by the wrapper has also been made optional and removed from the multiplex output.


jbeilstenedmands (approved):
Thanks, it's great that the initial cell clustering is now visible in the html 👍

jbeilstenedmands:
Note I think you can manually renumber the newsfrag to 940.feature, then the CI will be happy.

PR #942, opened 2026-07-31
xia2.ssx: Add 'wait_for_images' live-processing mode

ndevenish:
Allow a xia2.ssx job to be started at the beginning of a data collection and pause before each batch until that batch's image data has actually been written to disk, so each ~1000-image block starts processing as soon as it is available.

- Wait for the initial file to become available and readable (for launching before the main nxs file is ready).
- New PHIL scope wait_for_images { enable, method(*swmr|file_close), timeout, interval }, threaded through AlgorithmParams.
- wait_for_batch_images() polls dxtbx.nexus.get_available_frame_count() until every frame a batch needs (max global image index) is written.
- Gated at the top of process_batch (covers the serial and parallel paths) and in the cumulative assess-crystals / geometry-refinement loops for a cold start.
- On timeout the batch loop stops gracefully so data reduction still runs on the images collected so far.

Default-off: the standard pipeline is unchanged.


jbeilstenedmands (approved):
Thanks, I'm happy that this is well gated from the default code paths so it is quite safe to include, the code itself looks sensible.
The parameters being expert levels 3 also means this is somewhat hidden.

PR #945, opened 2026-08-25
xia2.ssx: Use removesuffix rather than rstrip for outputs

ndevenish:
str.rstrip(".mtz") strips any trailing '.', 'm', 't' or 'z' characters rather than the suffix, so record_merge_files could truncate the names it records for the merged mtz, log, json and html files. e.g.

`dials.merge.first.html ➡️ dials.merge.firs.html`

This isn't currently an issue, but could be with future features.


jbeilstenedmands (approved):
Thanks Nick, good spot.

Issue #944, opened 2026-08-14
xia2.ssx - allow passing of category names for series repeat


jbeilstenedmands:
Currently, `xia2.ssx` has the `dose_series_repeat=N` option, and it will label the outputs as `dose_1.mtz`, `dose_2.mtz` etc.

It would be nice to be able to provide names for the different groupings, to be able to generalize/customize this, e.g.
`series_repeat=first second last`, where the names would be passed through to the files and the repeat size determined from the length of the list of names. Should only require minor changes.

PR #948, opened 2026-08-26
xia2.ssx: Add series_repeat option for naming series groups (#944)

ndevenish:
series_repeat= generalises dose_series_repeat= by taking a list of names rather than a count: the repeat size is the number of names given, and the names label the merged output files and directories in place of the dose_1, dose_2, ... naming.

Names may be given comma-separated or as a quoted, space-separated list, as phil does not split multi-word values. A name may be given more than once, in which case each occurrence is numbered in order, padded to the digits needed for that name, i.e. series_repeat=dose,dose,apo gives dose_1, dose_2 and apo.

Completes #944.


jbeilstenedmands:
Just a comment, the naming might introduce a slight inconsistency: `dose_series_repeat` takes an integer, while `series_repeat` takes a list of strings. However there is also `dials.split_still_data`, which is a standalone program to splits data into the groups (e.g. if you wanted to separate integrated data before running data reduction), whose main parameter is `series_repeat` which takes an integer. Although I think it's only me and a few beamline scientists that use `dials.split_still_data`, and I can't think of an obviously better named and concise alternative parameter for this PR.

jbeilstenedmands:
Agreed solution is to allow `series_repeat` to be either a single integer or a list of names for the different groups, and implement this for both `xia2.ssx` and `dials.split_still_data`. Currently, `dials.split_still_data` uses the generic name of `group_0`, `group_1` etc., so that could be the default name in `xia2.ssx` if using an integer for `series_repeat`.

PR #949, opened 2026-09-02
Add command for analysing pixel countrate

pblowey:
Photon counting detectors such as the Dectris Eiger2, employ countrate corrections to account for the under-reporting of counts caused by deadtime. The manufacturers define a maximum trusted range, which corresponds to the range that the manufacturer deems these countrate corrections to be valid and values that exceed this range are returned as invalid pixels.

It has been reported that, towards the upper limit of the trusted range, the countrate corrections can be inaccurate, which has an impact on data processing, particularly data processing involving anomalous diffraction data.

This PR adds "xia2.countrate" as a command to process diffraction images, analyse whether the data is likely to be affected by poor countrate corrections and to suggest a maximum transmission at which the data could be collected in a good count-rate correction regime.

xia2.countrate comprises running dials.import and dials.find_spots, creating a histogram of the pixels found in the strong reflection shoeboxes and then analysising the pixel intensities as a proportion of the maximum trusted range of the detector to determine whether the data is at risk. Linear scaling is then used to propose a new maximum transmission at which a reference percentile of the pixel intensities can be scaled to a target percentage of the trusted range. 

Default parameters of the 99.9th percentile and 10% of the detector trusted range are selected, meaning that a transmission that should keep at least 99.9% of pixels below 10% of the detector trusted range will be suggested. At 10% of the detector trusted range, the actual vs corrected countrate is close to linear and, subsequently, the reported counts are more accurate. The 99.9th percentile was selected as this represents a large majority of pixels but in typical datasets on strongly diffracting crystal taken on I04 at DLS, still only corresponds to <30 % of the detector trusted range. This means that the reference data used to scale the tranmission is less likely to be affected by poor countrate corrections itself.

### DIFF
--- a/src/xia2/cli/countrate.py
+++ b/src/xia2/cli/countrate.py
@@ -1,0 +1,295 @@
+"""
+xia2.countrate: Process crystallography data through import and spotfinding.
+
+This program performs the following steps:
+1. Runs dials.import to import image data
+2. Runs dials.find_spots to find strong spots
+3. Performs additional processing on the spotfinding results
+
+Usage examples:
+    xia2.countrate image=/path/to/data/data_master.h5
+    xia2.countrate template=/path/to/data/image_####.cbf
+    xia2.countrate directory=/path/to/data/
+    xia2.countrate image=/path/to/data.h5 spotfinder.filter.min_spot_size=3
+"""
+
+from __future__ import annotations
+
+import logging
+import pathlib
+import subprocess
+import sys
+import traceback
+
+import iotbx.phil
+from dials.array_family import flex
+from dials.util.options import ArgumentParser
+from dxtbx.serialize import load
+
+import xia2.Handlers.Streams
+from xia2.Applications.xia2_main import write_citations
+from xia2.Handlers.Citations import Citations
+
+logger = logging.getLogger("xia2.cli.countrate")
+
+phil_scope = iotbx.phil.parse(
+    """
+input {
+    image = None
+        .type = str
+        .help = "Path to image file (e.g., data_master.h5)"
+    template = None
+        .type = str
+        .help = "Image template (e.g., image_####.cbf)"
+    directory = None
+        .type = str
+        .help = "Directory containing image files"
+}
+
+spotfinder {
+    min_spot_size = 2
+        .type = int
+        .help = "Minimum spot size in pixels"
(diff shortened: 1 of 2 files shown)


graeme-winter:
I have an interest in this ;-) 

However also

3. Analyses pixel intensities from shoeboxes to generate histogram of pixel intensities

This feels like something which would be a useful `dials.countrate` type tool as well?

I will watch the PR with interest!

src/xia2/cli/countrate.py
```diff
@@ -0,0 +1,349 @@
...
+
+This program performs the following steps:
+1. Runs dials.import to import image data
+2. Runs dials.find_spots to find strong spots
+3. Analyses pixel intensities from shoeboxes to generate histogram of pixel intensities
+
+Usage examples:
+    xia2.countrate /path/to/data/data_master.h5
+    xia2.countrate /path/to/data/image_####.cbf
+    xia2.countrate image=/path/to/data.h5 spotfinder.filter.min_spot_size=3
```
jbeilstenedmands, line 36: It looks like you don't yet have the functionality to pass through spotfinding arguments. But this can usually be achieved by adding the spotfinding phil scope into your program phil scope, with e.g. "include scope dials.algorithms.spot_finding.factory.phil_scope" as a line in your phil scope.

jbeilstenedmands:
From a first review this looks good overall.

I am in agreement with Graeme, it would be good to have the countrate calculations after spotfinding as a standalone program within dials, then this pipeline can run import, find_spots and countrate.

When you are happy with the code, please also include some tests as part of the PR(s), unit tests of individual functions are welcome and can go with the code in dials. Then a program/pipeline test using some of the regression data - we have data available in dials_data (pip install dials-data), you can see some example regression tests in `xia2/tests/regression`

pblowey:
Ok, thanks both for the feedback. I think its a good suggestion to make the countrate analysis a dials tool, so I'll work on that.

pblowey:
I've made a draft PR for "dials.countrate" (https://github.com/dials/dials/pull/3270) - I still need to add tests for this. I've also had a go at fixing the issue of dials.import not reading in the experimental transmission for nxmx format files in these PRs - https://github.com/cctbx/dxtbx/pull/896 and https://github.com/dials/nxmx/pull/78

PR #950, opened 2026-09-16
make sure small molecule output triggered in xia2.multiplex_filtering

amyjaynethompson:
It was found that xia2.multiplex_filtering would not output small molecule files, even if these options were set in the working-phil of the parent job. This is because the logic in `ScaleAndMerge.py` that sets the dummy composition from `small_molecule=True` is done in the `__init__` method, which is not touched by `xia2.multiplex_filtering`. Adding this logic into the static filtering method fixes this problem.


jbeilstenedmands (approved):
Thanks, I think you will need to manually rename the newsfragment.

